PR #1, opened 2019-07-18
Import Main Repo

MaxHillebrand:
This is a "blind" import of all the current documentation files I could find in the `WalletWasabi` and `Meta` repositories. I did not change anything of these files, I simply copied them in this new `WasabiDoc` repository.

Next we can work on restructuring the documentation with all the process tracked in this git repo. 

I suggest to not yet delete the original files, let's make sure that everything is done properly in the new repo first.


benthecarman:
This repo should have a different `README.md` than the wallet imo

MaxHillebrand:
Yes, agreed, the README will be changed as soon as this PR is merged and I can start editing in this doc.

I do think that there should be a chapter on developer documentation - it can be at the end of the list so it will only be found by those looking, but I think this is the right place to put it.

PR #5, opened 2019-07-18
add gitbook necessary structure

thunder-B:
This is really just to kickstart the docs and allow GitBook to generate the site. I included 5 basic files in there, but we can easily add more or remove all but the README if you guys prefer.

Note that the files needed for GitBook to generate the website need to be in the `./docs/` directory, as this is the one we'll designate to GitBook as the "root" of the docs.

### DIFF
--- a/docs/.gitbook.yaml
+++ b/docs/.gitbook.yaml
@@ -1,0 +1,5 @@
+root: ./docs/
+
+structure:
+  readme: ./README.md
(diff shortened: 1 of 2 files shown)


MaxHillebrand:
Ok, I can merge PRs now.
Thanks for your contribution @thunder-B!!! :heart:

PR #22, opened 2019-07-25
List and Describe the supported BIPs

MaxHillebrand:
This document will list all the BIPs and other industry standards that are supported by Wasabi Wallet already today, and what is planned on being supported in #twoweeks.


benthecarman:
For bips wasabi currently supports:
21

For bips wasabi will support in 2 weeks:
322

Maybe add a another doc or onto this one, ones that wasabi explicitly does not support:
69
70
37
125

MaxHillebrand:
Thanks Ben!

I think, now we have all the important BIPs listed + HWI. Are there any other industry standards [like HWI] that are supported? 

Regardless, next step is to give a tl;dr for each of these and maybe link to a wiki or something for further reference.

nopara73:
I think there's about 20-50 bips Wasabi supports or uses in some way through NBitcoin. Some BIPs here: https://github.com/MetacoSA/NBitcoin#description

docs/BIP.md
```diff
@@ -0,0 +1,44 @@
...
+
+### [BIP 158: Compact Block Filters for Light Clients](https://github.com/bitcoin/bips/blob/master/bip-0158.mediawiki)
+
+Block Filters (BIP 158) are the reverse - the client will connect to a Bitcoin node and say "Hey, for every block, I would like a condensed list of addresses that were affected." What would happen next is that a Bitcoin node would give the same filter that it gives to every client, because the client has thus far revealed nothing! Once a block filter has come in and the client believes that there is a transaction that affects the client, the client pings a single random node for a single full block. It then parses the block, and finds the transaction. This has been proven to be by far the best way to do light clients privately, and is the way Wasabi works today.
```
benthecarman: > Block Filters (BIP 158) are the reverse

Do "Block Filters (BIP 158) are the reverse of BIP 37", so if someone just clicks to 158 they know what you are talking about without needing to read 37

MaxHillebrand: Ah, yes, good catch!

MaxHillebrand:
Thanks @nopara73 for the reference.
I added the missing BIPs [SegWit] I could find in the NBitcoin readme.
I could not find another NBitcoin document that lists the other supported BIPs. @NicolasDorier, can you please point me in the right direction?

NicolasDorier:
I think the list is complete

MaxHillebrand:
Ok, then I'll merge this, and we can add explanations to with individual PRs.

PR #30, opened 2019-07-29
Different Sidebar for Each Section

thunderBiscuit:
This PR is mostly about reconfiguring the `config.toml` file to create the 4 different sidebars for each section (`Why Wasabi`, `Using Wasabi`, `Developing Wasabi`, and `FAQ`).

You will note that I did not want to make this PR about the decision of "what to include" in each category, and so it's simply a skeleton. Each "pillar" has an intro page (the page you land on when you click on the link in the navbar) and a few sections that currently don't have names other than "Section 1", "Section 2". Those are all easily configurable in the `config.toml` file.

Notice also that the navbar might need a bit of frontend love, because the GitHub link has a different left margin than the other links. I also find the other links a bit close together? Maybe a bit of breathing room would be nice. But this is all for another PR.


MaxHillebrand:
Please consider the restructuring to the three pillars described in #31 and somewhat implemented in #32. I think your PR is without merge conflicts in this regard. however, I did break the startup page with 654d55e2aafb687a797b498d2d6aa44166cd65b7, my bad, sorry :cry:

thunderBiscuit:
No merge conflicts. This PR sets the stage for further customization by creating a clean `config.toml` file further PRs can modify and work with. Note that the sections will need to be named, and the files can now be integrated in their proper place.

PR #36, opened 2019-07-30
More on how to build website locally

thunderBiscuit:
This PR basically adds to the section on how to build the site locally for development in the root `README.md` file. I also pushed the section at the bottom, below "how to contribute".

One message I wanted to get across is that contributors do not need to know how to use VuePress to contribute to the contents of the docs at all; they can simply PR changes to the markdown files and the website will be updated automatically. On the other hand, frontend devs who would like to help with the look should know how and where to go, and so I referred them to the vuepress documentation. (Note that there is a 1.x version of the vuepress docs and a 0.x version, and for the first 2 weeks I was working with the wrong docs, which caused me a lot of grief; if you are working with the vuepress docs, make sure you use the 1.0 version!)

Let me know if my explanations are clear. This is a bit more of a finesse section, and I am not sure if I hit the nail on the head just yet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,20 @@ If you are looking for the developers repository, it can be found here [WalletWa
 
 </br>
 
+## Contribute to the Documentation
+
+Just like the Wasabi wallet software, this documentation is libre and open source under the [MIT license](LICENSE). Not only do we appreciate the contributoin of peers, but we desperately need it! If you have a question, [open an issue](https://github.com/zkSNACKs/WasabiDoc/issues) and if you know the answer, then [make a pull request](https://github.com/zkSNACKs/WasabiDoc/pulls).
+
+The documentation has three main pillars, each of them have a separate folder for proper structure: [Why Wasabi](https://github.com/zkSNACKs/WasabiDoc/tree/master/docs/why-wasabi), [Using Wasabi](https://github.com/zkSNACKs/WasabiDoc/tree/master/docs/using-wasabi), and [Building Wasabi](https://github.com/zkSNACKs/WasabiDoc/tree/master/docs/building-wasabi), as well as an independent [Frequently Asked Question](https://github.com/zkSNACKs/WasabiDoc/tree/master/docs/FAQ). Please consider this repsoitory structure for hyperlinks, and use relative links whenever possible, for example `[What is a CoinJoin?](FAQ-Introduction.md#what-is-a-coinjoin)`.
+
+The graphical documentation is built with [VuePress](https://v1.vuepress.vuejs.org/), and the sidebar navigation can be [edited here](https://github.com/zkSNACKs/WasabiDoc/blob/master/docs/.vuepress/config.toml).
+
+If you have front end skills and would like to help improve the theme/look of the website, we welcome your input! Check out the [VuePress documentation](https://v1.vuepress.vuejs.org/) for details on how it works.
+
+On the other hand, if you'd like to contribute to the content of the site, know that you do not need to understand/deal with VuePress at all! You can make PRs pertaining to the markdown files in the `/docs/` directory, and if accepted, the merged changes will automatically be integrated into the documentation website by VuePress.
+
+<br>
+
 ## Build the Documentation Locally
 The Wasabi documentation website is built using an open source static site generator called VuePress. It is extrememly customizable and offers great defaults out-of-the-box. Check out the [VuePress documentation](https://v1.vuepress.vuejs.org/) for details on how to use it powerfully.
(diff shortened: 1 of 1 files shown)


MaxHillebrand:
Good re-write, thanks!

Though I'd like to get @dennisreimann feedback on the different install instructions, either with `npm` or `yarn`. I'm not sure how he does the deployment for the real thing, and it should be adequately reflected here.

dennisreimann:
I think this makes it too complicated to attract people interested in contributing. I don't understand how presenting alternative approaches for the setup process would help us, when the one we got has the minimum prerequisite and dependency count we can offer.

Even though I prefer using yarn in personal projects, I would highly discourage introducing it here:
It is just one more prerequisite and offers no additional benefits for this case.

Also:
- Why do we propose to install vuepress globally when installing it as a project dependency works  just fine?
- Why do we propose to use vuepress commands directly when we can hide implementation details (like which folders to use) with `npm start` and `npm run build`?

MaxHillebrand:
Ok, agreed with @dennisreimann - due to complexity, concept NACK to add the `yarn` install instructions. `npm` works very well and seems to be the right solution.

Regardless, @thunderBiscuit, I like your other write up. Can you please revert back to the `npm` instructions, and I'll merge the rest - Thanks!!

thunderBiscuit:
I see your points guys; there might not be much additional benefit. I am happy to remove. I'll still combine the text a bit and remove the explanations for how to deploy, how does that sound? I find the deployment stuff is more for the maintainers of the repo rather than the contributors to the documentation.

As for it being confusing however, I assume that if you get to this point on the page and are considering deploying locally you would be not thrown off too bad by the 2 possible well known solutions. Yarn is also the package manager recommended by VuePress (quote from the docs: _It is currently recommended to use Yarn instead of npm when installing VuePress into an existing project that has webpack 3.x as a dependency. Npm fails to generate the correct dependency tree in this case._) and I heard nopara say he didn't like npm in the dev call, so I just figured I'd offer the option. But in any case I'm not actually sure what webpacks do/are, so I don't think it matters in our case. If ever it's a problem, we can cross that bridge when we get there!

Revisions incoming.

MaxHillebrand:
Thanks @thunderBiscuit!

Quick question, why are the `npm run deploy` and `npm run build` commands removed?

dennisreimann:
@thunderBiscuit we are not using webpack as a dependency here. nopara mentioned he does not like the whole Node.js stack altogether.

thunderBiscuit:
@dennisreimann agreed we are not using it, so I'm totally down for just removing the yarn stuff (see latest commit, let me know if that's what you have in mind). I also think that if serious experienced developers come in and want to use yarn they will already know how to use it, so it's not like this section has them in mind I guess. It's more for the "slightly more new to contributing" contributor that the section is written, and in that way I can see why having one simple way to do it makes it easy for everyone.

@MaxHillebrand the two commands `npm run deploy` and `npm run build` are basically a cool trick to hide/simplify the true vuepress commands being run in your terminal when you use them. Check out the `scripts` object in the `package.json` file:
```
  "scripts": {
    "start": "vuepress dev docs",
    "build": "vuepress build docs",
    "deploy": "./deploy.sh"
  },
```

This is where Dennis defines the behaviour of `npm run <command>`. So for example when you do `npm run start`, the command line actually executes `vuepress dev docs`. You could run either and the result is the same. This behaviour is customizable, so it makes it powerful because you can embed more complicated scripts that are still executed in one clean command, making it easy for the users.

The reason I removed them from the readme is that those last two scripts (`build` and `deploy`) are really just for you and nopara and whoever is maintaining the repo. Our contributors coming in will never have to run those, and indeed they would not even work for them, (I think Dennis had them in there for you and I and the few people who would be building the site at first). For example `npm run deploy` will attempt to run the `deploy.sh` script, which pushes to the main repo on line 17. It would return an error for everyone else other than you and nopara because no one else has the required credentials, so I just think it might confuse people (especially if they think they somehow need to run it to propose their changes or something like that). On the other hand they will be necessary for you every time you want to update the site (until we have an automatic build set up), in which case you would run first the `build` and then the `deploy`, and the site would be updated on GitPages. But yeah those two are really just for you.

dennisreimann:
> The reason I removed them from the readme is that those last two scripts (build and deploy) are really just for you and nopara and whoever is maintaining the repo. Our contributors coming in will never have to run those, and indeed they would not even work for them,

> so I just think it might confuse people (especially if they think they somehow need to run it to propose their changes or something like that).

@thunderBiscuit Good point!

PR #43, opened 2019-07-31
Import Images to New Structure

### LINKED ISSUE #14: Add image directory
We already have some Wasabi info-graphics, and will most certainly create more. Further, we will have several screenshots of Wasabi in the documentation.

Thus we need to implement the infrastructure to properly store the images in the repo and display the images in GitBook. 

I'm not sure what is best practice, maybe @thunderBiscuit knows?

MaxHillebrand:
This is an import of all the images I could find in the incumbent repository, where previously we used Imgur to host the pictures. This PR puts a `.png` of each of these pictures in the directory `WasabiDoc/docs/.vuepress/public`. With this, we no longer rely on a third party to host the images. Because the new documentation is not part of the main Wallet, I don't see an issue with increasing the data size of the repository. It also seems that VuePress loads faster when it does not have to pull from Imgur.

In the markdown file where the picture should be shown, it can be referenced with ``, and VuePress will automatically display the image in html. This relative link is suggested in the [VuePress documentation](https://vuepress.vuejs.org/guide/assets.html#asset-handling).

Please note that this PR is a blind import of the existing images and the establishment of the proper directory structure. Many of the images are out of date, some should be removed, and many more should be added. But I think it's best to do this in a separate PR.

Closes #14.


dennisreimann:
Looks good for me. Agreed, hosting the images ourselves will be more performant than using a third-party service for that.

PR #53, opened 2019-08-02
Adapt theme from new website design

### LINKED ISSUE #4: Make Wasabi fitting theme
According to the [VuePress documentation](https://vuepress.vuejs.org/default-theme-config/) we can configure the layout and theme of the documentation very thoroughly. So I suggest that we make it fitting to the new [WasabiWallet.io](https://WasabiWallet.io) theme.

Same [logo](https://github.com/zkSNACKs/Meta/blob/master/WasabiPressKit.zip), colors, font, etc.

We can also consider if it should look similar like the application itself, with that dark mode [yes plz sir!] and blue highlights. 

I remember that there were efforts to change the website, and I'm not sure what the latest is on this... This might change some of the branding, and we can easily adapt to this.

[Edited, now we're using VuePress, not GitBook]

dennisreimann:
Proposal on how the design of the upcoming webiste could be adapted for the docs. 

I deployed the changes so that everyone can take a look here: 
https://docs.wasabiwallet.io/

@mayankchhabra @MaxHillebrand @thunderBiscuit Looking forward to your feedback!

If merged, this would close #4.

### DIFF
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -2,7 +2,8 @@ module.exports = {
   title: "Wasabi Documentation",
   description: "Wasabi is an open-source, non-custodial, privacy focused Bitcoin wallet for desktop.",
   head: [
-    ["link", { rel: "icon", href: "/favicon.ico" }]
+    ["link", { rel: "icon", href: "/favicon.ico" }],
+    ["link", { rel: "stylesheet", href: "https://fonts.googleapis.com/css?family=Inconsolata:400,700|Playfair+Display:700&display=swap" }]
   ],
   plugins: [
     "@vuepress/back-to-top"
--- a/docs/.vuepress/styles/index.styl
+++ b/docs/.vuepress/styles/index.styl
@@ -1,5 +1,20 @@
 // Custom styles, see this file for details:
 // https://github.com/vuejs/vuepress/blob/master/packages/docs/docs/miscellaneous/design-concepts.md#reasonable-priority-management
+
+// Default is dark mode: @media (prefers-color-scheme: dark)
+// Overrides can be done using @media (prefers-color-scheme: light)
+html
+body
+  background $darkBgColor
+  color $lightTextColor
+
+  @media (prefers-color-scheme: light)
+    background: #fff
+    color $textColor
+
+    .theme-default-content code
+      color $codeTextColor
+
 #app
   // fonts
   font-family Inconsolata, monospace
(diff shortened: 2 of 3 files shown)


mayankchhabra:
Rock 'n roll! Great work @dennisreimann 👏👏

I like the approach of overwriting the styles with the main website's colors and typography as from whatever I researched, creating a custom Bootstrap layout for vuepress would have been a lot of pain, plus no future updates from vuepress in terms of layout and features. However, the main website uses a lot of bootstrap classes such as `text-uppercase/lowercase`, `colors`, and `spacings`, so you can try pulling those classes to achieve similar results. 

Other than that, everything looks great. Agree with Max that a dark mode would be much easier on the eyes as the docs are pretty text-heavy, so you can try inverting the colors. 

Some small suggestions that you can consider to achieve a more cohesive theme: 

Screen Shot 2019-08-03 at 5 14 07 pm

_Changing the "Wasabi Documentation" font to Playfair Display, and the blue button to the green with uppercase text, and I think we should also change the text to "Read the docs" instead of "reclaim your privacy"._

***

Screen Shot 2019-08-03 at 5 13 28 pm

_The website uses lowercase for all the headings, so we can do the same in the docs as well for all the headings in Playfair font_

***

Screen Shot 2019-08-03 at 5 13 05 pm

_Uppercase Inconsolata looks great for sub-headings like this one "Manual public key import", with a tad bit of increased letter spacings._

dennisreimann:
> Can you please increase the font size a little?

Done with 712cf978e06f335f8c016c6b6e4fe1244d4f9f71.

@mayankchhabra I also incorporated your suggestions and feedback in the finetuning commit.

You can take a look at the current website, I deployed it with the recent changes including dark mode… https://docs.wasabiwallet.io/

Speaking of dark mode: According to the [quantitative research done by @MaxHillebrand](https://twitter.com/HillebrandMax/status/1157554875611398144) our users prefer the dark side, so we made it the default. But as we don't want to put off the users that would rather use light mode, we cater to theirneeds with the [CSS prefers-color-scheme media query](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme).
Browsers that support this show the theme depending on the users system preferences, see [my tweet](https://twitter.com/DennisReimann/status/1157714320324747265) for details.

MaxHillebrand:
As written in #57, we should increase the font size of the top nav bar.

Maybe even with the general font as well?

dennisreimann:
Thanks for the kind words, Max! I hope this will be one of many little contributions that will make the whole more than the sum of all parts. The efforts of the community help grow this project to become a valuable resource about Wasabi and privacy in Bitcoin in general. A much needed and very valuable resource for generations to come 😀

PR #67, opened 2019-08-04
Change all "you" to "You"

MaxHillebrand:
This PR changes the capitalization of `you` so that all variants are upper case `You`. The master has both types, and we should be consistent. I did not find any other instance of a lower case `you`.

The question is if we want to introduce this writing convention.
- If yes, then merge this PR.
- If no, then close this PR and open a new one to lower case all upper case `You`


thunderBiscuit:
I'm not a super fan of the `You` version. But is it something that's been like a Wasabi vibe since the beginning?

MaxHillebrand:
No, the incumbent version is confused and inconsistent with `You` and `you`.

We need to have a clear and consistent rule which one of the two to use.

I like the `You`, in German when a pronoun is capitalized, then it is a token of respect - thus when we capitalize the `You`, then we give our respect to the individual user.

yahiheb:
I think we should go with lower case `you`. It feels strange to see `You` capitalized everywhere.

MaxHillebrand:
Yes, agreed - I noticed this especially when writing, lower case `you` is the way to go.

I'm closing this PR, and will open a new one that makes it lower case on all accounts.

PR #79, opened 2019-08-05
Add details toggles for FAQs

### LINKED ISSUE #72: [FEATURE] drop-down for FAQ
Is there a way to add a drop down text box?

This might be useful for the FAQ. All questions are displayed in a single line, each has a drop down arrow, and when clicked it reveals the answer to the question. Somewhat similar to how the [new wasabiwallet.io](https://www.weareblock.space/wasabi/) does the FAQ. This would make the `[[toc]]` obsolete.

dennisreimann:
As requested in #72 this introduces the custom block `::: details` which can be used to initially hide long-form content and first present a summary. In this case the FAQ leverage it to better structure the questions and answers and provide a better overview than the previously used table of contents.

Thoughts: Even though I really like the output, my concern about this is that we need to use a custom block for that and the authoring of the markdown does not seem intuitive. IMHO the headlines plus plaragraphs feel more natural. Feedback welcome!

If merged, this would close #72.

Signed-off-by: Dennis Reimann <mail@dennisreimann.de>

### DIFF
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -15,24 +15,29 @@ module.exports = {
       type: "details",
       render (tokens, idx) {
         const token = tokens[idx]
-
+        // turn details headline into summary
         if (token.type === 'container_details_open') {
           const next = tokens[idx + 1]
           const match = token.info.trim().match(/^details\s+(.*)$/)
           let title = match && match[1]
           if (next.type === 'heading_open' && !title) {
-            const headNext = tokens[idx + 2]
-            title = headNext && headNext.content || ''
+            const headContent = tokens[idx + 2]
+            const headClose = tokens[idx + 3]
+            // hide headline and its contents
+            next.hidden = headClose.hidden = headContent.hidden = true
+            headContent.children = []
+            // extract title
+            title = headContent.content || ''
           } else {
             title = ''
           }
           const slug = slugify(title)
-          return `<details><summary><h3 id="${slug}"><a href="#${slug}" aria-hidden="true" class="header-anchor">#</a> ${title}</h3></summary>`
+          return `<details id="${slug}"><summary><a href="#${slug}" aria-hidden="true" class="header-anchor">#</a> ${title}</summary>`
         } else if (token.type === 'container_details_close') {
           return '</details>'
         }
       }
-    }],
+    }]
   ],
   markdown: {
     extendMarkdown (md) {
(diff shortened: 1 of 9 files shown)


MaxHillebrand:
Ok, I think that I now have all the FAQ properly integrated. A couple things that I noticed:

- The question is no longer shown in the sidebar as `###` header. I think it would be good to have it in here, and when clicked, to jump to the question and open it.
- when there is a text box `:::danger:::` in the question, then this messes with the drop down, it closes right after this text box.

dennisreimann:
All of that makes me wonder if this is a good approach. From the semantics of the output this would be the right solution, but it has too many tradeoffs and makes things fragile on the authoring side, which I don't like.

I also thought about not hiding the HTML and exposing it on the authoring side (i.e. using `<details><summary>` directly in the markdown), but that feels very unnatural and is brittle too (for instance teh markdown inside those blocks does not get rendered anymore).

All in all I think we should settle this.

dennisreimann:
> when there is a text box :::danger::: in the question, then this messes with the drop down, it closes right after this text box.

Regarding nesting of custom blocks: [outer block start/end require more `:`](https://github.com/markdown-it/markdown-it-container/issues/6#issuecomment-213789283).

dennisreimann:
Ok, I got it worked out. Please take another look @MaxHillebrand and let me know what you think :)

dennisreimann:
Ok, last revision, better than before, but stil not perfect: The sidebar hightlight gets tripped up and does not work well with toggling containers.
Improved the authoring side with normal headline handling, though.

I am leaving it to you Max, whether or not this should get merged. I am still not really happy with this, but would not invest any more time here, as this feels very fragile and would require upstream changes in VuePress to handle this properly.

MaxHillebrand:
Ok, tested, it works decent: 
- drop down works
- headers are shown in sidebar
- hyper link to question works

One thing that might be further improved
- when linking to specific question, drop down the content automatically

I really really really like the concept - I think it's much better than incumbent FAQ.

But I am concerned that you don't think it's a robust architecture... I guess that what we have now in this PR is "good enough", and if it requires a little more work to get it even better, then I think this might be worthy the investment of time.

Also, we might consider to communicate with the VuePress team to get them tinkering on this as well - maybe even offer them a bounty as thanks for their awesome work so far. [We have the budget of 1k USD per month, their tool has helped us a lot, so I'd like to consider them anyhow]

So I give concept ACK, but kick the issue back to you @dennisreimann in regards to implementation ACK, if you are comfortable enough with this PR, let's merge it!

dennisreimann:
Ok, I got it working, so please check again, @MaxHillebrand.

In case you think the current state is mergeable we still need to rebase it with your latest changes though.

dennisreimann:
Updated the branch to match the latest master. Please review and decide whether or not to merge before updating more FAQs :)

PR #80, opened 2019-08-06
Edit link of FAQ navigation bar

yahiheb:


yahiheb:
Maybe we should then do the same with `Using Wasabi` and `Building Wasabi`.

PR #82, opened 2019-08-06
Edit title and description

yahiheb:


docs/.vuepress/config.js
```diff
@@ -2,8 +2,8 @@ const customBlock = require('markdown-it-custom-block')
 const youtubeEmbed = path => `<div class="ytEmbed"><iframe src="https://www.youtube-nocookie.com/embed/${path}" frameborder="0" allow="autoplay;encrypted-media;picture-in-picture" allowfullscreen></iframe></div>`
 
 module.exports = {
-  title: "Wasabi",
-  description: "Wasabi is an open-source, non-custodial, privacy focused Bitcoin wallet for desktop.",
+  title: "Wasabi docs",
+  description: "Wasabi Wallet is an open-source, non-custodial, desktop, privacy focused Bitcoin wallet.",
```
MaxHillebrand: I actually like the "for desktop" at the end here.

this makes the beginning a nice 3 argument punch "open-source" "non-custodial" "privacy-focused" which is also the tagline of the start page.

yahiheb: Maybe `desktop` should be after the 3 argument punch. Often we don't find `an application for desktop` but `a desktop application` is more common.
I also I think the title should be `Wasabi docs` and in the description we should say `Wasabi Wallet`.

PR #88, opened 2019-08-06
Remove License file in docs folder

yahiheb:


MaxHillebrand:
But why?
I'd like to show the license in the docs as well.

yahiheb:
But isn't that redundant? Why do we need it in the docs folder?

MaxHillebrand:
I mean - it should be shown in the docs website that Wasabi is MIT license, otherwise it's only in the github repo.

MaxHillebrand:
I've fixed that the license was missing in the side bar.
closing this now :)

yahiheb:
But can't you use the other license file?

MaxHillebrand:
no, because it's not in the /docs/ repo...

PR #89, opened 2019-08-07
Edit the main README file

yahiheb:
Correct some typos, remove some redundant sentences, put new sentence in a new line and fix some links.

PR #96, opened 2019-08-07
Link to git repo (README)

yahiheb:


MaxHillebrand:
NACK, in this case, the link is for readers of the contribution checklist, not for editors of this specific file. 
Thus I think it's better to link to the GUI website with pretty dark mode :smiley:

yahiheb:
@MaxHillebrand Then in this [commit](https://github.com/zkSNACKs/WasabiDoc/pull/89/commits/ef71903c936c55a040fd42a47bb9d24c54cadeae) where you said:

> the links in the main README should be to the git repo, because this is for contributors, not readers.

we should link to the docs because people are there to read about the pillars of the documentation not to edit specific files.

MaxHillebrand:
Yes, in the general case, for example the links to the [main pillars](https://github.com/zkSNACKs/WasabiDoc#the-pillars-of-the-documentation), they go to the git, because here the reader wants to work on these documents.

But if the reader of the main README has a question about how to contribute, then he's looking for the live website, and not how to edit the file that tells him how to contribute.

yahiheb:
I am not sure that the reader of the main pillars wants to work on these documents.
I think because this is the documentation repository so mostly people are hear to read and learn not to work (they might be willing to work though).
So IMO most of the links should link to the docs website.

MaxHillebrand:
Good point...
Our repo structure is very confusing, I think that when we link inside the `/docs/` in the main README, then this can help to clear up how this docs is structured in general.

idk what the best practice is here, I don't have a strong opinion about either, I just want to try to make it less confusing...

PR #98, opened 2019-08-08
Links on FAQ main page

benthecarman:


MaxHillebrand:
We decided against having manual table of contents, since they are a pita to maintain, and because we have automatic toc in the sidebars and with the `[[toc]]` markdown command.

So instead, I'd rather like to have a general explanation what this FAQ is, what problem it solves and why it is useful. 

Not sure exactly what to write about here, that's why I was hesitant to put anything in these README files...

So, concept NACK for this toc PR, we would have to delete this when we finally write the proper introduction.

PR #147, opened 2019-08-12
FIX FAQ-UseWasabi.md

RiccardoMasutti:
Some FAQs were not displayed.

### DIFF
--- a/docs/FAQ/FAQ-UseWasabi.md
+++ b/docs/FAQ/FAQ-UseWasabi.md
@@ -35,23 +35,23 @@ You have successfully setup your wallet when you click `I wrote down my Recovery
 ![](/WalletManagerRecoveryWords.png)
 :::
 
-::::details
+:::details
 ### What password should I use?
 The password you set is used to encrypt the mnemonic recovery words, and the private keys stored on the computer.
 This is the password that will unlock your bitcoin to anyone who has access to the backup or computer.
 You will need to type in the password before you can spend from Wasabi wallet.
 So if your backup gets compromised, this password is the only thing protecting your precious sats.
 
-:::danger
+::danger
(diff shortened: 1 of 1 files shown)


MaxHillebrand:
We have here one box inside the other. The outer box is the `details` dropdown for the FAQ. The inner box is the `danger` colored warning box for text highlights.

According to [this](https://github.com/markdown-it/markdown-it-container/issues/6#issuecomment-213789283) coding convention, you should add one additional `:` to the outer `details` box, this should have `::::`. The inner `danger` box should remain at three `:::`.

As far as I remember, the local build with the pre-PR version was working properly. So, at least how I understand it, your PR breaks this, and it would show the text below the `danger` box outside the dropdown. But currently I have some issues to build master, thus I cannot check, will do so asap.

How do you experience the bug of having questions not showing? In your local build, or on the main website? If on the main website, then the issue that the questions are not showing is because we still do not have Azur pipelines setup properly... So we need to manually deploy after every commit, and I think the last deployment was a day ago, so all the commits since are not on the `docs.wasabiwallet.io` site.

MaxHillebrand:
OK, I just checked with the local master build, it works here, the questions that you want to edit in this PR are all shown properly already. I also checked your PR, and you would break it, because it is confused what the inner and outer box is. I'll add a better explanation in the [writing conventions](https://github.com/zkSNACKs/WasabiDoc#frequently-asked-questions).

For writing on the docs, it might be nice to use the [local build](https://github.com/zkSNACKs/WasabiDoc#build-the-documentation-locally), it is easy to setup, and it gives you a local preview of your edits before you even have to do a git commit. This is great for testing and playing around. 

I hope we soon in #twoweeks have the Azure pipeline setup working, then all the git commits will be instantly deployed on the live website.

Issue #150, opened 2019-08-13
[BUG] embeded videos eat RAM


MaxHillebrand:
### Problem

PR #73 adds the plugin to embed youtube videos and load it when the markdown file is loaded in VuePress. Files like the `/using-wasabi/InstallPackage.md` or the `/FAQ/FAQ-usingwasabi.md` have many videos embedded. When loading them either locally or over website, it eats all the RAM of my virtual machine, and it completely crashes. When I remove all the embedded videos and rebuild, everything works fine.

### Solution

As discussed with @dennisreimann in PM, it might be possible to only pre-load the thumbnail, and load the full embedded video only when it is clicked.

PR #152, opened 2019-08-14
Videos: Add preview, play on click

dennisreimann:
This PR addresses the problem described in #150.

Bezond that it offers UX and performance improvements by showing a preview image of the video first and loading the video only on click. This also conforms to the GDPR consent as YouTube is only connected in case the user opts-in to watching the video.


dennisreimann:
We can add an overlay for the preview image that indicates the play button. Have already though about that, just didn‘t find the time to do so. I‘ll add it.

PR #153, opened 2019-08-15
Add OSX video guide

nopara73:
- Removed the "check out..." part, because it made no grammatical sense in this case and it'd be consistent with how we did in the Windows part.
- Added OSX guide by oceans4all (https://twitter.com/oceans4all/status/1160894308205023233)


dennisreimann:
The Debian/Ubuntu part also has the „check out“ intro. Let‘s remove this too. 

Other than that this looks good!

PR #156, opened 2019-08-20
Update FAQ-Introduction.md

RiccardoMasutti:
Answer to minimal requirements. As discussed privately with @nopara73.
It's constantly changing, there may be more.


MaxHillebrand:
What is the minimum required for Linux?

nopara73:
Linux 2.3.8 😄

nopara73:
Just a joke. I have no idea. Specifying minimum requirements also means we must guarantee that it'd work properly if those are met. But the problem is you can't guarantee anything on Linux. Use Ubuntu or live with the potential issues.

PR #161, opened 2019-08-21
[STRUCTURE] make BIPs dropdown

MaxHillebrand:
This changes the layout of `using-wasabi/BIPs.md`.

- Every BIP is a dropdown `:::details:::` box
- The link to BIP github is in the text of the box
- The explanation of the BIP is in the text of the box
- If the explanation of one BIP references another BIP, then it is an internal link `###`
- The automatic toc is removed, instead manual links only to `##`

### DIFF
--- a/docs/using-wasabi/BIP.md
+++ b/docs/using-wasabi/BIP.md
@@ -22,7 +22,6 @@ Here is a list of all the supported and integrated Bitcoin Improvement Proposals
 ### BIP 21: URI Scheme
 
 [BIP 21: URI Scheme](https://github.com/bitcoin/bips/blob/master/bip-0021.mediawiki)
-foo
 :::
 
 :::details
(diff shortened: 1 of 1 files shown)


docs/using-wasabi/BIP.md
```diff
@@ -16,103 +18,193 @@ Here is a list of all the supported and integrated Bitcoin Improvement Proposals
...
+:::details
+### BIP 72: bitcoin: uri extensions for Payment Protocol
+
+[BIP 72: bitcoin: uri extensions for Payment Protocol](https://github.com/bitcoin/bips/blob/master/bip-0072.mediawiki)
+:::
+
+:::details
+### BIP 84: Derivation scheme for P2WPKH Based Accounts
+
+[BIP 84](BIP.md#bip-84-derivation-scheme-for-p2wpkh-based-accounts) defines a standard derivation scheme for hirarchical deterministic wallets BIP 32, specifically for segregated witness P2WPKH [BIP 173](BIP.md#bip-173-base32-address-format-for-native-v0-16-witness-outputs).
```
yahiheb: for `hierarchical`

docs/using-wasabi/BIP.md
```diff
@@ -16,103 +18,193 @@ Here is a list of all the supported and integrated Bitcoin Improvement Proposals
...
+This allows to generate one root master seed that can derive a tree of public keys with different paths [BIP 44](BIP.md#bip-44-multi-account-hierarchy-for-deterministic-wallets).
 `m / purpose' / coin_type' / account' / change / address_index`.
 Wasabi specifically uses this standard `m/84'/0'/0'`.
 On the TestNet and on the RegTest Wasabi deviates from the standard and usees `m/84'/0'/0'` instead of `m/84'/1'/0'`.
```
yahiheb: and `uses`

docs/using-wasabi/BIP.md
```diff
@@ -16,103 +18,193 @@ Here is a list of all the supported and integrated Bitcoin Improvement Proposals
...
+[BIP Tapscript](https://github.com/sipa/bips/blob/bip-schnorr/bip-tapscript.mediawiki) defines a slight variation on Bitcoin’s existing Script language to be used in [BIP Taproot](BIP.md#bip-taproot) merkle spends.
+:::
+
+
+## What is not supported
+
+:::details
+### BIP 37: Connection Bloom Filtering
+
+[Bloom filters](BIP.md#bip-37-connection-bloom-filters) are filters that a client will send a Bitcoin full node which says "Hey, if you see any transactions that get caught in this filter, they may or may not be mine!".
```
yahiheb: send `to` a Bitcoin full ... in `these filters`

dennisreimann:
IMHO it overcomplicated things when burying single links inside the details container. Are we going to describe each item or are there BIPs where we will just link to? For the latter I would try to come up with a better solution.

MaxHillebrand:
Agreed @dennisreimann, it is a bit awkward when the link to the BIP is the only thing in the drop down box... But I do want to have a description of all these BIPs, at bare minimum the abstract of each, and then I think this drop down is better than before.

PR #162, opened 2019-08-21
Rename BIP.md file to BIPs.md

yahiheb:
If this gets merged I will open a PR [here](https://github.com/nvk/wallets-recovery) to update the link.

### DIFF
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -123,7 +123,7 @@ module.exports = {
             "/using-wasabi/WalletGeneration.md",
             "/using-wasabi/PasswordFinder.md",
             "/using-wasabi/Daemon.md",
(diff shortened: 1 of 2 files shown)


MaxHillebrand:
It might be that `SupportedBIPs.md` is an even better name for this file. 
Your PR is not up to date, #161 changes the BIP structure.
And I already spoke with NVK, he updated his list already. I also would like to have a step by step guide how to do it with Electrum, in another `using-wasabi/WalletRecovery.md` file.

yahiheb:
I think that `SupportedBIPs.md` is not a better name because in that file we have also information about not supported BIPs.
I will close this PR because it has many conflicts and open a new one.

> I also would like to have a step by step guide how to do it with Electrum, in another using-wasabi/WalletRecovery.md file.

That would be great.

PR #163, opened 2019-08-21
Rename BIP.md file to BIPs.md

yahiheb:


MaxHillebrand:
Did you check if there are any other references in the docs that use the old link?

yahiheb:
I checked some but not all.

PR #127, opened 2019-08-09
Edit title and description

yahiheb:
Second attempt. Read [comment](https://github.com/zkSNACKs/WasabiDoc/pull/82#discussion_r311049462).

PR #207, opened 2019-09-18
Edit title and description

yahiheb:
I think we should say wasabi docs on the main page of the docs website.

wasabi docs

### DIFF
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -15,8 +15,8 @@ const youtubeEmbed = (id, path) => `
 const themeColor = "#211b24"
 
 module.exports = {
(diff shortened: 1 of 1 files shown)


yahiheb:
YES. Finally I got it merged after 3 attempts :rofl:
#82 #127

MaxHillebrand:
Yes, hehe, I was giggling while merging, all good things come in threes...

I just looked through these two PRs, and for both of them my issue was not with `Wasabi Docs`, but with the other lines changed in the PR. So I learn that 1] I shall be more precise what I like and dislike when I close something, and 2] it's good to have separate PRs that change only one thing, like this one.

PR #210, opened 2019-09-21
Add tutorial about Wasabi on Tails

RiccardoMasutti:


PulpCattel:
If there's a part that you'd like to add directly into your doc, feel free to do so. 
If something need to be changed, feel free to ask and I'll try to make it better fit your request.
Wasabi + Tails is dope.

MaxHillebrand:
@PulpCattel would you mind to open a PR that creates a separate markdown file in the `/using-wasabi/` directory and import your guide? 
I don't want to copy your work and get the commit credit, especially because this is considered in the monthly contribution fund...

PR #224, opened 2019-09-27
Update status page link

yahiheb:
I think this should only be merged after https://github.com/zkSNACKs/WalletWasabi/pull/2355 is merged. @molnard?


nopara73:
ACK depending on what @molnard wants to do with PR https://github.com/zkSNACKs/WalletWasabi/pull/2355

PR #225, opened 2019-09-28
Small edits

yahiheb:


docs/FAQ/FAQ-UseWasabi.md
```diff
@@ -527,7 +527,7 @@ Average: 10 sat/vbyte: 1,680 satoshis per participant (0.0168 %)
 High: 100 sat/vbyte: 16,800 satoshis per participant (0.168 %)
 :::
 Here, we already observe that in some extreme cases, the miner is earning almost 0.2%!
-Recall that Wasabi has a coordinator fee that is capped at 0.3%, so in these extreme cases the actual fee paid by the user (you) is not 0.3% as promised, but closer to 0.5%.
+Recall that Wasabi has a coordinator fee that is capped at 0.3%, so in these extreme cases the total fee paid by the user (you) is closer to 0.5%.
```
MaxHillebrand: Is this actually correct?
I thought we are rather below 0.3%, and not above?

yahiheb: ~0.5% is the total fee (coordinator fee + mining fee).
~0.5% in this example comes from 0.3% coordinator fee + 0.168 % mining fee (High: 100 sat/vbyte: 16,800 satoshis per participant).

There are edge cases where you can pay even more than 1% (0.3% coordinator fee + some sats for mining fee + 0.7% if the change is too small).
@nopara73 please check if what I said is correct.

yahiheb: Also note that in https://docs.wasabiwallet.io/FAQ/FAQ-UseWasabi.html#what-are-the-fees-for-the-coinjoin we don't mention anything about the mining fee.
I think we should state in the beginning that the CoinJoin fee is a combination of two fees (coordinator fee + mining fee).

nopara73: Officially it's capped at 1%.

PR #226, opened 2019-10-01
Add Introduction about Cold Storage

RiccardoMasutti:


docs/using-wasabi/ColdWasabi.md
```diff
@@ -5,6 +5,129 @@
...
+- Lamination is prone or degradation over time and puncture or cuts that could allow moisture to get trapped in the paper and cause deterioration or rotting in some circumstances | store in cool dry place
+- Can burn or be smoke damaged
+- 'Fireproof' & 'Fire-resistant' boxes can help protect paper in a small house fire but be warned that they can sometimes fall apart in the fire and get wet if the fire is put out with water.
+Remember people can just carry out a small safe
+:::
+
+:::details
+### Engraved / etched/ ablated/ stamped on a piece of metal
+- Anyone who can see it, can steal it
+- Some metals can deteriorate or corrode, choose a good metal; also store your metal away from direct contact other metals. - Some metals that are corrosion resistant have low melting points, are extremely expensive, or hard to machine.
```
yahiheb:
```suggestion
- Some metals can deteriorate or corrode, choose a good metal; also store your metal away from direct contact other metals.
- Some metals that are corrosion resistant have low melting points, are extremely expensive, or hard to machine.
```

docs/using-wasabi/ColdWasabi.md
```diff
@@ -5,6 +5,129 @@
...
+- Metals can still deform or melt from heat, destroying any engraved SK. "Most house fires do not burn hotter than 1,200 degrees Fahrenheit.
```
yahiheb:
```suggestion
- Metals can still deform or melt from heat, destroying any engraved SK.
"Most house fires do not burn hotter than 1,200 degrees Fahrenheit.
```

docs/using-wasabi/ColdWasabi.md
```diff
@@ -5,6 +5,129 @@
...
+- Silver, gold, copper, brass, bronze, nickel, cobalt, would survive the housefire fire unmelted.
+Some Aluminium alloys can survive but you have to have the right ones.
+At around 1500° Steel and Nickel should be okay.
+Titanium is above the housefire range and so is tungsten, however tungsten rings are known to shatter due to the brittle nature of the very hard metal.
+:::
+
+:::details
+### Stored digitally on a computer
+- Computers can crash, making data recovery expensive
+- Data can still technically be recovered after a system is abandoned by the user. In some cases data can be recovered after multiple overwriting attempts and physical destruction (as long as the attacker can get all or most the pieces) so if you copy files to a new computer and ditch the old one, be careful.
```
yahiheb:
```suggestion
- Data can still technically be recovered after a system is abandoned by the user.
In some cases data can be recovered after multiple overwriting attempts and physical destruction (as long as the attacker can get all or most the pieces) so if you copy files to a new computer and ditch the old one, be careful.
```

docs/using-wasabi/ColdWasabi.md
```diff
@@ -5,6 +5,129 @@
...
+- Can burn or be smoke damaged
+- A traditional hard disc drive can have data corrupted by powerful magnetic fields and can physically shatter
+- A non-negligible amount of HDDs suffer from factory defects that will cause them to fail unexpectedly during their - lifetime
+- Accidents can happen that could result in loss of data
+- Solid state drives (SSDs) will lose data if unpowered, they may last years before this becomes a problem but it is unwise to store long-term data in unpowered SSDs
+- If connected to internet it is another attack vector and the safety is only as good as the encryption used; I don't know what I would recommend but it wouldn't be BitLocker. Someone could be trying to break into the computer constantly. Even with good encryption if the machine or location is compromised the key could be stolen as soon as it is decrypted.
```
yahiheb:
```suggestion
- If connected to internet it is another attack vector and the safety is only as good as the encryption used; I don't know what I would recommend but it wouldn't be BitLocker.
Someone could be trying to break into the computer constantly.
Even with good encryption if the machine or location is compromised the key could be stolen as soon as it is decrypted.
```

docs/using-wasabi/ColdWasabi.md
```diff
@@ -5,6 +5,129 @@
 }
 ---
 
+# What is Cold Storage
+
+Cold storage in the context of Bitcoin refers to keeping a reserve of Bitcoins offline.
```
yahiheb:
```suggestion
Cold storage in the context of Bitcoin refers to keeping a reserve of bitcoin offline.
```

docs/using-wasabi/ColdWasabi.md
```diff
@@ -5,6 +5,129 @@
...
+This is often a necessary security precaution, especially dealing with large amounts of Bitcoin.
```
yahiheb:
```suggestion
This is often a necessary security precaution, especially dealing with large amounts of bitcoin.
```

yahiheb:
the content of the PR is from https://en.bitcoin.it/wiki/Cold_storage and from a book titled '#SmartCustody' https://github.com/BlockchainCommons/SmartCustodyWhitePapers/blob/master/%23SmartCustody-_Simple_Self-Custody_Cold_Storage_Scenario.md
they are under CC BY 3.0 and CC-BY-SA-4.0 licences.
is there any problem using their content in our repo without giving them credit?
also I don't think it is a good idea to fill the Wasabi Docs with everything related to bitcoin.
we should stick to Wasabi related content IMO.
and if necessary we can simply link to trustworthy content.

RiccardoMasutti:
@yahiheb you are right, I forgot to mention source on the footer! Do you prefer to insert on the same page or right in /using-wasabi/credits or something like that?

yahiheb:
I really don't have any idea. Let's see what @MaxHillebrand says.

RiccardoMasutti:
> I really don't have any idea. Let's see what @MaxHillebrand says.

What about #246 ?

PR #240, opened 2019-10-07
Import RPC doc

MaxHillebrand:
this imports the documentation from [here](https://github.com/zkSNACKs/WalletWasabi/blob/393cea8ec21881e287429d4b204d3608f9760380/WalletWasabi.Documentation/RPC.md).
@lontivero @schulterklopfer if you would like, you can delete the old one, and from now on commit to this documentation.


lontivero:
@MaxHillebrand you can't imagine how happy makes me to know that you find my pet feature so valuable but given our time is finite we have to focused on privacy oriented features that do not add maintenance costs and this feature is not one of those.

However, we are going to reevaluate the addition of the rpc server after January 2020. How knows...

In case the server is merged this PR has to be merged too, of course. Until now, we cannot merge it.

PR #251, opened 2019-10-10
Create InstallBackend.md

RiccardoMasutti:
Add guide on how to install backend

### DIFF
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -134,7 +134,8 @@ module.exports = {
             "/using-wasabi/ColdWasabi.md",
             "/using-wasabi/LurkingWifeMode.md",
             "/using-wasabi/Daemon.md",
-            "/using-wasabi/BIPs.md"
+            "/using-wasabi/BIPs.md",
+            "/using-wasabi/InstallBackend.md"
           ]
         },
     {
--- a/docs/FAQ/FAQ-Installation.md
+++ b/docs/FAQ/FAQ-Installation.md
@@ -203,4 +203,8 @@ You may want to start with `--help`.
 
 ## Further Questions
 
-- How do I install the Wasabi backend server?
+:::details
+## How do I install the Wasabi backend server?
+
+To install Wasabi backend server you can follow [this guide](/using-wasabi/InstallBackend.md).
+:::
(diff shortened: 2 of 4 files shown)


nopara73:
My position did not change, but I am not the maintainer of this repo, so only take my comment as an advice. The backend deployment docs is maintained in the main repo: https://github.com/zkSNACKs/WalletWasabi/blob/master/WalletWasabi.Documentation/BackendDeployment.md

This document is in the main repo, because it is crucial for our development, and must fall under strict scrutiny.  

This document regardless if it's correct in its current form or not, it will not be maintained.  
Btw, I have similar maintenance concerns of many other information in this repo. It's just a matter of time that this documentation start to be a disservice instead of a service by spreading outdated, therefore incorrect information. I suggest trying to tackle the maintenance issue systemically before accepting yet another unmaintainable document.

Maybe put big warnings to pages those are poised to be outdated, this way you can accept much more things. Although I'd personally just make sure that the information can stand the test of time, but this repo isn't seem to be created with that in mind.

yahiheb:
I agree with Nopara, I think the backend deployment docs should be in the main repo.
Why not simply link to it.

RiccardoMasutti:
> I agree with Nopara, I think the backend deployment docs should be in the main repo.
> Why not simply link to it.

I made this PR because there was an unanswered FAQ [here](https://docs.wasabiwallet.io/FAQ/FAQ-Installation.html#further-questions).

If you prefer, we can simply link the guide instead of adding it inside docs but in my opinion would be awesome to have this content inside docs.
Or, even better, put the entire guide in docs but, as @nopara73 suggests, add a big warning on top of it specifying to follow the link to the main repository for the constantly updated guide

yahiheb:
> If you prefer, we can simply link the guide instead of adding it inside docs

IMO that is the way to go.

MaxHillebrand:
how about we delete the content of the main repo docs, and only have this one here, and when ever you need to change the backend deployment, then you work on it in this repo?

molnard:
I agree with nopara. I would not even mention backend installation in the docs because it could start a Lavina and might encourage users to try. Docs is a user guide. Users do not want to install backend. 
- Installing backend contains so many details and failure points that it should be done only by experts.

If somebody wants to install a backend, contact us or try it alone with all the responsibility.

PR #265, opened 2019-10-22
Edit dust threshold image

yahiheb:
Should be merged only after https://github.com/zkSNACKs/WalletWasabi/pull/2454 is merged.


nopara73:
Please explain. In this PR you removed the image file, but in PR https://github.com/zkSNACKs/WalletWasabi/pull/2454/ you did not touch it.

What is the connection?

yahiheb:
In this PR I simply edited the picture to show `0.00001` instead of `0.0001` if the other PR is merged.

nopara73:
Aaah, this is the WasabiDoc repo. I missed that. I thought you modified an icon in the Wasabi repo or something. Anyhow, still NACKing `0.00001` :p

nopara73:
Btw now it's misleading. You cannot change the dust limit of existing dust limit of users', so it'll stay 0.0001 for everyone who ever used Wasabi. It will just be lowered for new users.

yahiheb:
I don't understand.

nopara73:
The dust limit is stored in the config file. You only changed the default dust limit, the old default dust limit is already set for all existing users, so in your PR you only changed new users' dust limits.

yahiheb:
That is correct.
BTW this is the doc repo

yahiheb:
Do you want to change the dust limit for existing users?
I am not sure we should do that, maybe some users have their dust threshold under `0.00005`

nopara73:
I don't know what is the right thing to do. Spreading misinformation is the worst IMO. There are a few things those I have in mind.  

- Make sure I explain both the old dust limit and the new one.
- I remove the dust limit discussion altogether to make the article evergreen.
- I talk about the dust limit as a variable, not as a concrete number, and maybe when I would have to refer to the dust limit explicitly like in the illustration, I'd refer to it as it'd be 1BTC so it'd be clear for everyone that's not the real dust limit. This way the article can be kept evergreen.

PR #268, opened 2019-10-25
Add currentVersion variable

dennisreimann:
This substitutes the `${currentVersion}` with whatever is defined in the variables.js file. The goal is to have a single place to maintain the current Wasabi version number as it is used in multiple places. This could also be extended for other cases, but I'm putting it out here to discuss the idea.

~Note: I haven't found a way to make the substitution work in fenced code blocks yet. that's why this commit uses the `<pre><code>` fallback for code blocks on the Deterministic Builds page. It's ugly, nevertheless worth it imho.~

Feedback appreciated!

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,7 +133,6 @@ bar
 :::
 ```
 
-
 ### Frequently asked questions
 
 Use this markdown box with the headers to write the FAQ:
@@ -159,6 +158,12 @@ answer answer answer.
 :::
 ::::
 ```
+
+### Variables
+
+To have a single place to maintain universal strings like the current Wasabi version number, we use variables in the Markdown (i.e.  `${currentVersion}`).
+These variables are managed in `docs/.vuepress/variables.js`.
+Occurrences of `${variableName}` get substituted before the Markdown is processed.
 
 ### Thanks goes to the Wasabikas, Osu!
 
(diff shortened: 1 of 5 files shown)


molnard:
utACK

well done!

yahiheb:
That is a very good job.
I tested it and it works fine but look how it looks in fenced code blocks:
Capture1

@molnard can you confirm that those are all the places where the version should be updated.

dennisreimann:
@yahiheb good catch! I didn‘t check the light theme – this is nasty thought, as I had to replace the standard code blocks with custom ones, that‘s why it‘s not visible. Will fix this – thanks for spotting that!

RiccardoMasutti:
> @yahiheb good catch! I didn‘t check the light theme – this is nasty thought, as I had to replace the standard code blocks with custom ones, that‘s why it‘s not visible. Will fix this – thanks for spotting that!

I will fix #269 too.

dennisreimann:
I found a better way to achieve this. It also has the benefit of being able to keep the fenced code blocks – so the constraint mentioned originally is gone.

@yahiheb This also automatically fixes the issue with the light theme.

@RiccardoMasutti Can you incorporate the update and change the `{{ $page.currentVersion }}` occurrences in your PR to `${currentVersion}`? Thanks!

🙏

yahiheb:
That is great. One thing to note that in the github files you see ` ${currentVersion}` instead of the version.
tACK.

dennisreimann:
Yeah, but we cannot prevent that I think. However – it‘s kinda readable and reveals the intention 😄

PR #269, opened 2019-10-25
Substitute X.X.X version with variable

RiccardoMasutti:
Following #268


dennisreimann:
I left those in on purpose because I wasn‘t sure whether or not we should be abstract in these cases. This can be discussed further here – in case we decide to merge I‘d like to clean this up a bit, as it‘s kind of a hack to use the `<pre><code>` workaround. Don‘t really like that and would try to find a better way before spreading this further.

yahiheb:
As @dennisreimann said this shouldn't be merged until #268 is merged.

yahiheb:
This is now ready to merge after #268

PR #289, opened 2019-11-02
Add FAQ about spent coin status

RiccardoMasutti:


MaxHillebrand:
And for the record, this is a fork of @nopara73 FAQ repo https://github.com/nopara73/NoparaFaq, I think it's better to have this explanation in the docs, so to have it all in one place.

yahiheb:
@MaxHillebrand I think @nopara73 wanted to use the text editor to change that on purpose.

And by changing that the next sentence `Wasabi will think that the wallet was loaded in...` becomes wrong (or unnecessary)

MaxHillebrand:
@nopara73 is there any reason why the manual config file change? In my experience it is fixed with the GUI way too...

nopara73:
> @nopara73 is there any reason why the manual config file change? In my experience it is fixed with the GUI way too...

There is none.

yahiheb:
One thing to note though, nopara suggested to edit the `yourwalletname.json` file but using the GUI the `Config.json` file will be the one edited, is there any difference?

PR #306, opened 2019-11-09
Use PGP public key consistently

yahiheb:
Use `PGP public key` instead of `key` or `PGP key` or `public key`.


dennisreimann:
Might also be a use case for a specific [variable](https://github.com/zkSNACKs/WasabiDoc#variables).

yahiheb:
> Might also be a use case for a specific variable.

But this will not change over time.

dennisreimann:
> But this will not change over time.
Correct, but this way we would ensure that there's a single source of truth.

dennisreimann:
I totally missed that this is primarily a wording change and not about the PGP key itself. Sorry for that! (Nevertheless we might want to use a variable for the key itself).

PR #311, opened 2019-11-10
Update with data folder link

RiccardoMasutti:


yahiheb:
This is wrong.
`Program Files\WasabiWallet` is not the same as `Wasabi data folder`.

PR #315, opened 2019-11-10
Remove #twoweeks from docs

RiccardoMasutti:
Not everyone knows #twoweeks joke and may create disappointments and incorrect information (especially to newbies)

### DIFF
--- a/docs/FAQ/FAQ-Contribution.md
+++ b/docs/FAQ/FAQ-Contribution.md
@@ -55,7 +55,7 @@ There are three developers who have the ability to merge code into master branch
 Wasabi is far from complete, there are many Wasabikas contributing every day to make this tool of self defense even more powerful.
 Because Wasabi is libre and open source software, anyone can support the project without asking for permission.
 Thus it is relatively tricky to give a precise roadmap with what will be implemented in the near and distant future.
-You can check the [ToDo list](/building-wasabi/ToDo.md) for a somewhat up-to-date summary of the short term priorities of the next #twoweeks, and what might be implemented in the distant future.
+You can check the [ToDo list](/building-wasabi/ToDo.md) for a somewhat up-to-date summary of the short term priorities of the next two weeks, and what might be implemented in the distant future.
 :::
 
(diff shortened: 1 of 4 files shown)


docs/FAQ/FAQ-Contribution.md
```diff
@@ -55,7 +55,7 @@ There are three developers who have the ability to merge code into master branch
 Wasabi is far from complete, there are many Wasabikas contributing every day to make this tool of self defense even more powerful.
 Because Wasabi is libre and open source software, anyone can support the project without asking for permission.
 Thus it is relatively tricky to give a precise roadmap with what will be implemented in the near and distant future.
-You can check the [ToDo list](/building-wasabi/ToDo.md) for a somewhat up-to-date summary of the short term priorities of the next #twoweeks, and what might be implemented in the distant future.
+You can check the [ToDo list](/building-wasabi/ToDo.md) for a somewhat up-to-date summary of the short term priorities of the next two weeks, and what might be implemented in the distant future.
```
yahiheb: Using `two weeks` here is worse than using `#twoweeks` because you are precising the date and this is wrong.

docs/FAQ/FAQ-Installation.md
```diff
@@ -152,7 +152,7 @@ Yes, just follow the [Wasabi Setup on Tails](/using-wasabi/WasabiSetupTails.md)
 :::details
 ### How do I compile Wasabi from source?
 
-A new version of Wasabi is released when ready, roughly once every #twoweeks.
+A new version of Wasabi is released when ready, roughly once every two weeks.
```
yahiheb: NACK. A new version is not released every two weeks.

MaxHillebrand:
I NACK, the #twoweeks is a fun inside joke. It might not be suitable for a "serious" documentation, but I think it's fun here... 
if we agree to remove the fun, then we should re-write the sections properly, but having it "two weeks" and "the future" doesn't make much sense, it removes the fun, and then it looks serious, but is factually wrong....

Issue #309, opened 2019-11-10
Docs site unavailable: 404 File not found


PulpCattel:
The site is unavailable.

![Docs 404](https://user-images.githubusercontent.com/51127079/68542636-05d18a00-03af-11ea-8d19-fbe073f81974.png)

MaxHillebrand:
Yes, #310 fixed it!

PR #307, opened 2019-11-09
Add const variable for zkSNACKs public key fingerprint

yahiheb:

PR #264, opened 2019-10-21
Improve Cold Storage Guide

6102bitcoin:

PR #317, opened 2019-11-11
DONT MERGE testing breaking it

MaxHillebrand:
This is a test that re-implements the bug introduced in #307, reported in #309 and fixed in #310. 
I am testing the newly added requirement that the azure pipeline test is successful before being able to merge. This PR should fail the test, and thus I should not be able to merge it. Let's see...

### DIFF
--- a/docs/.vuepress/variables.js
+++ b/docs/.vuepress/variables.js
@@ -1,6 +1,6 @@
 // Add custom variables here:
 const variables = {
-  currentVersion: '1.1.9.2',
+  currentVersion: '1.1.9.2'
   zksnacksPublicKeyFingerprint: '6FB3 872B 5D42 292F 5992 0797 8563 4832 8949 861E'
 }
 


MaxHillebrand:
While the checks are running, I cannot merge the branch.
screenshot_20191111_095804

When the checks fail, I cannot merge the branch.
screenshot_20191111_100007

Here are the logs of the failed checks.
screenshot_20191111_100107

However, as an administrator, I can still merge the branch, but it is a big red button, so it's very scary looking.

This is very good, hopefully we will now catch such a mistake earlier.

MaxHillebrand:
There are however PRs that fail the test, but they seem to be working properly... That's when I can use admin privileges to merge anyhow, after being sure that everything works actually.

#289 #276 #264 etc...

molnard:
Sometimes CI tests fail because of random errors, in that case u can re-run them. 
image
Click on the red x and login to azure and re-run.

PR #319, opened 2019-11-11
Redirect to BIPs.md

RiccardoMasutti:


yahiheb:
> we should link to the internal explanation of the BIPs instead, they have a link to the external source.

I disagree.

MaxHillebrand:
I should explain my rational here...
The [BIPs](https://github.com/zkSNACKs/WasabiDoc/blob/master/docs/using-wasabi/BIPs.md) file has a full list of all the BIPs wasabi supports, or chooses not to support. Each of them is linked to the external `/bitcoin/bitcoin` repo where the original pieces are. As well, there is an explanation of what the BIP is, and why it is [not] supported.

I would prefer that every where in the documentation, we link to our BIPs file, because here the user can get a tl;dr that is tailor made for wasabi. And he can easily find the external link to it. Also, we keep the user within the docs, and he can see other content like the additional BIPs more easily.

What do you say about my rational? 
I'm not married to the idea, but I do think it is better than always linking to the external repo without any approachable explanation.

yahiheb:
> As well, there is an explanation of what the BIP is, and why it is [not] supported.

In the [what is supported today section](https://docs.wasabiwallet.io/using-wasabi/BIPs.html#what-is-supported-today) only BIP 84 and BIP 158 have an explanation of what the BIP is.
Also each BIP has an abstract explaining what it is, so I don't think we need to write another explanation, except when there is something to add specific to Wasabi.

MaxHillebrand:
Well, I think often the BIPs are a bit confusing, especially for n00bs, and users might get overwhelmed by them... So a quick tl;dr is what I envision here.

But well, this is out of the scope of this PR...

How should we agree on the linking to BIPs.md or to external?

yahiheb:
> Well, I think often the BIPs are a bit confusing, especially for n00bs, and users might get overwhelmed by them... So a quick tl;dr is what I envision here.

I agree with that

RiccardoMasutti:
> Well, I think often the BIPs are a bit confusing, especially for n00bs, and users might get overwhelmed by them... So a quick tl;dr is what I envision here.
> 
> But well, this is out of the scope of this PR...
> 
> How should we agree on the linking to BIPs.md or to external?

Ok let's redirect users to BIPs.md. Let me update pages referring to bips

RiccardoMasutti:
@MaxHillebrand @yahiheb Fixed everything. Ready to merge.

yahiheb:
I still think we should link to the external source.

RiccardoMasutti:
> I still think we should link to the external source.

I linked everything to BIPs.md. Every BIP inside that page contains a link to the full document on bitcoin repo.

yahiheb:
You did a great job Riccardo, but the problem is we send the readers to one place and then we send them to the external place without real benefit. Why not send them directly to the external source.

RiccardoMasutti:
@yahiheb I think the best thing would be redirect people to BIPs.md which contains a summary.
Every BIP inside BIPs.md then contains a link to the complete doc

PR #326, opened 2019-11-18
Update Windows Version

RiccardoMasutti:
> Our dependencies are ending Windows 7 support the same day as Microsoft ends Windows 7 extended support: https://home.bt.com/tech-gadgets/computing/windows-7/windows-7-support-end-11364081315419
> 
> So we have no choice but to end the support in January 14, 2020, too. Considering this, fixing this issue could be a waste of time. Sorry about it.


MaxHillebrand:
Since we still have a couple weeks until it is no longer supported, should we keep it now, and rather add that support for Windows 7 will be depreciated soon - then update the doc in January again?

PR #346, opened 2019-11-25
[BUILDING] Add summary of license

MaxHillebrand:
This is a copy of the [GitHub summary](https://github.com/zkSNACKs/WasabiDoc/blob/master/LICENSE) of the MIT license. It's a nice and short explanation of what is later explained in a bit of lawyer speak. Also the table makes everything easy to understand at one glance.

I also fixed the one line per sentence writing convention.

Ready for review and merge. I didn't change any wording, so it should be a no-brainer.

### DIFF
--- a/docs/building-wasabi/LICENSE.md
+++ b/docs/building-wasabi/LICENSE.md
@@ -13,7 +13,6 @@ Wasabi Wallet and Wasabi documentation is licensed under the MIT License.
 
 A short and simple permissive license with conditions only requiring preservation of copyright and license notices.
 Licensed works, modifications, and larger works may be distributed under different terms and without source code.
-Permissions
 
 | Permissions    | Limitations  | Conditions                   |
 |----------------|--------------|------------------------------|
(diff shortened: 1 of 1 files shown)


docs/building-wasabi/LICENSE.md
```diff
@@ -9,22 +9,24 @@
...
+Licensed works, modifications, and larger works may be distributed under different terms and without source code.
+
+| Permissions    | Limitations  | Conditions                   |
+|----------------|--------------|------------------------------|
+| Commercial use | Liability    | License and copyright notice |
+| Modification   | Warranty     |                              |
+| Distribution   |              |                              |
+| Private use    |              |                              |
+
+
```
yahiheb: I think this section should be moved above the `# MIT License` title because it is not part of it.

MaxHillebrand: but `# MIT Lisense` is the title of the document, which is always first.
i do get your point though... not sure what to do

PR #350, opened 2019-11-25
Add FAQ about difference between Linux versions

RiccardoMasutti:


yahiheb:
I honestly think we shouldn't include this kind of questions in the docs.
If someone is willing to install Wasabi on a Linux distribution and they don't know the difference (I don't know the difference myself) or which package is for them they better do some research, or ask somewhere but we can't put every answer in the FAQ.

PR #367, opened 2019-12-02
Added a new "Scenarios" section to list all the main practical scenarios that a user could face

SimoneBovi:
This is only a draft for now but that's the thing that I personally would have really liked the first time I approached Wasabi and Bitcoin privacy.

To me the Scenarios section should contain a list of all the main practical scenario that a user could face. It is something that maybe could be redundant with other documentation already available but I think that wrapping information this way could be really useful for someone new to these topics by serving as a quite precise and concrete summary.

Let me know what you think and feel free to suggest new scenarios or improvements to the ones I have already committed.

If you don't like the idea no problem, this would mean I just spent some time to clear my mind, so it is all good in any case :)

### DIFF
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -103,6 +103,10 @@ module.exports = {
         link: "/building-wasabi/"
       },
       {
+        text: "Scenarios",
+        link: "/scenarios/"
+      },
+      {
         text: "FAQ",
         link: "/FAQ/"
       },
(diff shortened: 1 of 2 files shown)


RiccardoMasutti:
The concept is interesting but much of the content is repeated in other sections.

I believe this PR needs a major overhaul.
What do you think @MaxHillebrand @yahiheb ?

PR #381, opened 2019-12-07
Add CoinJoin examples

RiccardoMasutti:


RiccardoMasutti:
> Good idea!
> 
> Why have you chosen these specific coin joins, just because they're the same on the website?
> 
> should we also add some additional information, like `First ever 100 anonset CoinJoin`?

Yes, they're the same

PR #382, opened 2019-12-07
Improve the explanation of label edits

RiccardoMasutti:


PulpCattel:
Isn't enough to change the wallet.json file to change the label of every address? Do you want to omit that because it's too technical?

MaxHillebrand:
Well @PulpCattel this is kinda hacky, and it messes up with the smart coin selection algorithm that uses labels. So we need a proper solution instead.

PR #341, opened 2019-11-22
Add example of good labeling

RiccardoMasutti:

PR #384, opened 2019-12-07
Add label examples

RiccardoMasutti:


yahiheb:
We removed those examples here https://github.com/zkSNACKs/WasabiDoc/pull/341/files
I don't think putting the amount and the date is really useful.
What we can do is maybe say this `You can read more about labels [here](/FAQ/FAQ-UseWasabi.md#why-do-i-have-to-label-my-address)`
Also there is this https://github.com/zkSNACKs/WasabiDoc/pull/341#issuecomment-557540790

RiccardoMasutti:
> We removed those examples here https://github.com/zkSNACKs/WasabiDoc/pull/341/files
> I don't think putting the amount and the date is really useful.
> What we can do is maybe say this `You can read more about labels [here](/FAQ/FAQ-UseWasabi.md#why-do-i-have-to-label-my-address)`
> Also there is this [#341 (comment)](https://github.com/zkSNACKs/WasabiDoc/pull/341#issuecomment-557540790)

I thought of the same thing, but then we risk bumping the end user from one side to the other...

yahiheb:
Still the example is not good IMO.

PR #390, opened 2019-12-12
Fix Tails tutorial capitalization and add v1.1.10

RiccardoMasutti:


RiccardoMasutti:
> Good.
> 
> The open question is if bitcoind can be run in persistent storage of Tails?

Yes it's possible

MaxHillebrand:
But you would need > 300GB USB stick... 
Might be good to mention this in the tip box.

Issue #145, opened 2019-08-12
What do you need to get the keys?


yahiheb:
I would love an explanation of [this](https://github.com/zkSNACKs/WasabiDoc/blob/master/docs/using-wasabi/WalletGeneration.md#what-do-you-need-to-get-the-keys). I do not really get it.
The title says `What do you need to get the keys?` but reading that do not answer that question.

MaxHillebrand:
Yes, this is only somewhat of a placeholder based on [this](https://github.com/zkSNACKs/WalletWasabi/issues/2049#issuecomment-520192355) very good knowledge by @lontivero - I just didn't have time to write it out properly...

Lucas, if you have some spare time, maybe you can write up some more of the details? [not your priority though!]

lontivero:
> What do you need to get the keys?

In that section I list all the possible pieces of data that can be used to recover a HD wallet. In other words: all the ways with can compute the (master) extended key.

The details about how to get the privates keys from an extended key is specified in the bip32 and I am sure I cannot explain it better. https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki

yahiheb:
@MaxHillebrand From your list:
**Data**: entropy | **Does it need password to recover?** No, this is unencrypted
You don't need the password to decrypt it (because it is unencrypted) **but** you need the password to recover the wallet.
@lontivero Your list and Max's list are different.

MaxHillebrand:
closed by #396

yahiheb:
This is not solved yet.

PR #396, opened 2019-12-14
Correct sentence describing the password usage

yahiheb:
As discussed with @lontivero the password doesn't encrypt the mnemonic recovery words but rather it encrypts the private key of the extended private key, and the encrypted private key is stored on the computer.

### DIFF
--- a/docs/FAQ/FAQ-UseWasabi.md
+++ b/docs/FAQ/FAQ-UseWasabi.md
@@ -17,7 +17,7 @@
 You can generate as many new wallets as you'd like, for no extra cost and without asking for permission.
 Go to the `Wallet Manager` tab and the `Generate Wallet` menu.
 As with everything in Wasabi, you are required to label this new wallet. Make sure that you are precise so that you know what this is for later.
-The password is used to encrypt the private key on the computer and on the backup.
+The password is used to encrypt the private key (extracted from the extended private key) on the computer.
 **Careful!!**
 **Without knowledge of the password, you CANNOT spend your bitcoin!!**
 Make sure that you properly back up and write down this password.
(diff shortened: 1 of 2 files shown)


docs/FAQ/FAQ-UseWasabi.md
```diff
@@ -98,11 +96,11 @@ You can also easily reach it from inside Wasabi: `File > Open > Data Folder`.
 Although you can backup your private keys with the mnemonic words and password, this is only a last resort recovery.
 If you want to also secure your address labels, the anonset and additional metadata, then you can do a digital backup.
 Simply copy the `WalletBackups` folder with the `wallet.json` files from your [Wasabi data folder](/FAQ/FAQ-UseWasabi.md#where-can-i-find-the-wasabi-data-folder) onto suitable hardware, for example an encrypted USB stick.
-Note that this file has the encrypted private keys, meaning that you only need the password to spend the bitcoin.
+Note that this file has the encrypted private key (extracted from the extended private key), meaning that you only need the password to spend the bitcoin.
```
lontivero: Ok, this is correct but it is just a technical/implementation detail that can help a few people to understand better the process while confuse the majority of the users. What is the point to explain that the private key is the one from the extended private key? 

Imo I would just say "... has the encrypted private key", this is better than saying "private keys" what is simply not true.

yahiheb: > What is the point to explain that the private key is the one from the extended private key?

Most of the people know about a private key related to one address which is important, but the private key mentioned in this sentence is not the same one and is more important.
(Just a clarification, and if someone is willing to know they might do some research)

MaxHillebrand: I am in favor of accuracy, even if it comes at the expense of confusing some users... 
I don't like talking down to users, I assume they are willing enough to accumulate difficult but accurate knowledge and try to understand it. 

So I ACK these changes

RiccardoMasutti:
This should close #145 right? @yahiheb

yahiheb:
No it shouldn't.

PR #405, opened 2019-12-17
[FAQ] add wassabee explanation

MaxHillebrand:
A copy of nopara's explanation [here](https://github.com/zkSNACKs/WalletWasabi/issues/2278#issuecomment-531651480) with some minor changes.

### DIFF
--- a/docs/FAQ/FAQ-Installation.md
+++ b/docs/FAQ/FAQ-Installation.md
@@ -237,7 +237,7 @@ In doing so you will help users who use the same antivirus.
 
 The most obvious thing would be to call the executable `Wasabi Wallet.exe?  on Windows and `Wasabi Wallet` on Linux and Mac.
 
-However launching Wasabi Wallet from certain terminals/command lines are problematic, so we should remove the spacing: `WasabiWallet`, and the case sensitivity differences between the operating systems can also cause issues, so that must go too: `wasabiwallet`.
+However launching Wasabi Wallet from certain terminals/command lines is problematic, so we should remove the spacing: `WasabiWallet`, and the case sensitivity differences between the operating systems can also cause issues, so that must go too: `wasabiwallet`.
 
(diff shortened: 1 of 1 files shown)


docs/FAQ/FAQ-Installation.md
```diff
@@ -230,4 +230,18 @@ If you are still not sure, before starting Wasabi, you can check that it is safe
 Remember to report Wasabi as a safe software/false positive to your antivirus.
 In doing so you will help users who use the same antivirus.
 :::
+::::
+
+:::details
+### Why is the executable called wassabee?
+
+The most obvious thing would be to call the executable `Wasabi Wallet.exe?  on Windows and `Wasabi Wallet` on Linux and Mac.
```
yahiheb:
```suggestion
The most obvious thing would be to call the executable `Wasabi Wallet.exe?`  on Windows and `Wasabi Wallet` on Linux and Mac.
```

docs/FAQ/FAQ-Installation.md
```diff
@@ -230,4 +230,18 @@ If you are still not sure, before starting Wasabi, you can check that it is safe
...
+
+However launching Wasabi Wallet from certain terminals/command lines are problematic, so we should remove the spacing: `WasabiWallet`, and the case sensitivity differences between the operating systems can also cause issues, so that must go too: `wasabiwallet`.
```
yahiheb:
```suggestion
However launching Wasabi Wallet from certain terminals/command lines is problematic, so we should remove the spacing: `WasabiWallet`, and the case sensitivity differences between the operating systems can also cause issues, so that must go too: `wasabiwallet`.
```

docs/FAQ/FAQ-Installation.md
MaxHillebrand: actually remove the ?, I hit the wrong key...

PR #418, opened 2019-12-21
Add password to recovering a Wasabi wallet

yahiheb:


docs/using-wasabi/WalletRecovery.md
```diff
@@ -11,7 +11,7 @@ There are two different ways that you can recover an already used wallet in Wasa
 
 [[toc]]
 
-## Mnemonic Recovery Words
+## Mnemonic Recovery Words and Password
```
MaxHillebrand: This might break some links [as it changes the `#` in the url]
do you know if they are used somewhere in the docs?

yahiheb: I don't know it is used somewhere.

PR #427, opened 2019-12-22
[GLOSSARY] zero link

MaxHillebrand:
Adds explanation of zero link frangibility framework

### DIFF
--- a/docs/glossary/Glossary.md
+++ b/docs/glossary/Glossary.md
@@ -305,5 +305,5 @@ Wasabikas are essentially builders and users of Wasabi in general.
 ZeroLink is a framework to holistically design a privacy and fungibility setup for Bitcoin.
 This encompasses more than just a single CoinJoin transaction, but also includes network level privacy defends against third parties spying.
 
-Read more: [ZeroLink: the Bitcoin Fungibility Framework](https://github.com/nopara73/zerolink)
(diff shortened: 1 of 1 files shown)


docs/glossary/Glossary.md
```diff
@@ -298,3 +298,12 @@ Read more: [Technical Overview of Wasabi Wallet](/building-wasabi/TechnicalOverv
 
 Wasabikas are essentially builders and users of Wasabi in general.
 :::
+
+:::details
+### ZeroLink
+
+ZeroLink is a framework to holistically design a privacy and fungibility setup for Bitcoin.
```
yahiheb: Maybe use the word `completely` instead of `holistically`?

docs/glossary/Glossary.md
```diff
@@ -298,3 +298,12 @@ Read more: [Technical Overview of Wasabi Wallet](/building-wasabi/TechnicalOverv
...
+This encompasses more than just a single CoinJoin transaction, but also includes network level privacy defends against third parties spying.
```
```suggestion
This encompasses more than just a single CoinJoin transaction, but also includes network level privacy defense against third party spying.
```

docs/glossary/Glossary.md
MaxHillebrand: complete would suggest it's finished, while holistically says it is all encompassing.

PR #430, opened 2019-12-22
[GLOSSARY] change

MaxHillebrand:
Adds glossary of change coins

### DIFF
--- a/docs/glossary/Glossary.md
+++ b/docs/glossary/Glossary.md
@@ -23,7 +23,7 @@ Read more: [Wasabi Supported BIPs](/using-wasabi/BIPs.md)
 :::details
 ### Change
 
-When in a spending transaction the provided input coins are larger than the value send to the destination address, then the leftover change is send back to an address of the same wallet.
+When in a spending transaction the provided input coins are larger than the value sent to the destination address, then the leftover change is sent back to an address of the same wallet.
 CoinJoins create standardized equal value outputs, and the change of a CoinJoin can easily be linked to the input, thus they have 0 anonymity set.
 Read more: [Change coins](/using-wasabi/ChangeCoins.md)
 :::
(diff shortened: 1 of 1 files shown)


docs/glossary/Glossary.md
```diff
@@ -20,6 +20,14 @@ A design document for introducing features or information to Bitcoin.
 Read more: [Wasabi Supported BIPs](/using-wasabi/BIPs.md)
 :::
 
+:::details
+### Change
+
+When in a spending transaction the provided input coins are larger than the value send to the destination address, then the leftover change is send back to an address of the same wallet.
```
SimoneBovi: When in a spending transaction the provided input coins are larger than the value sent to the destination address, then the leftover change is sent back to an address of the same wallet.

```suggestion
When in a spending transaction the provided input coins are larger than the value sent to the destination address, then the leftover change is sent back to an address of the same wallet.
```

RiccardoMasutti:
NACK.

docs/glossary/Glossary.md
```diff
@@ -20,6 +20,14 @@ A design document for introducing features or information to Bitcoin.
 Read more: [Wasabi Supported BIPs](/using-wasabi/BIPs.md)
 :::
 
+:::details
+### Change
+
+When in a spending transaction the provided input coins are larger than the value sent to the destination address, then the leftover change is sent back to an address of the same wallet.
```
```suggestion
In a spending transaction where the provided input coins are larger than the value sent to the destination address, then the leftover change is sent to a new address of the same wallet.
```

docs/glossary/Glossary.md
```diff
@@ -20,6 +20,14 @@ A design document for introducing features or information to Bitcoin.
...
+CoinJoins create standardized equal value outputs, and the change of a CoinJoin can easily be linked to the input, thus they have 0 anonymity set.
```
yahiheb: I think this sentence should be removed from this definition.

MaxHillebrand: I was thinking the same... 
```suggestion
```

MaxHillebrand:
`Change` is mentioned many times in the docs, it even has it's own chapter because of the many privacy problems with it, so I think definitely it should be included.

PR #432, opened 2019-12-22
[GLOSSARY] segwit

MaxHillebrand:
adds glossary on segwit

### DIFF
--- a/docs/glossary/Glossary.md
+++ b/docs/glossary/Glossary.md
@@ -111,6 +111,14 @@ Read more: [How can I display the fee in satoshi per byte?](/FAQ/FAQ-UseWasabi.m
 :::
 
 :::details
+### SegWit
+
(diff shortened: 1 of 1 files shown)


RiccardoMasutti:
NACK.
Should be `Bech32` synonymous imho.

MaxHillebrand:
> Should be added as Bech32 synonymous. - @RiccardoMasutti 

No, bech32 =/ segwit. bech32 is an encoding standard, which is also used elsewhere. For example taproot addresses will also be bech32 encoded.

RiccardoMasutti:
> > Should be added as Bech32 synonymous. - @RiccardoMasutti
> 
> No, bech32 =/ segwit. bech32 is an encoding standard, which is also used elsewhere. For example taproot addresses will also be bech32 encoded.

Yes I agree with that, but I'm thinking like a newbie.

MaxHillebrand:
> I'm thinking like a newbie.

A newbie will like to have a correct two sentence explanation tl,dr of what segwit is...

PR #435, opened 2019-12-22
[GLOSSARY] public key

MaxHillebrand:
adds public key to glossary

### DIFF
--- a/docs/glossary/Glossary.md
+++ b/docs/glossary/Glossary.md
@@ -100,7 +100,7 @@ Read more: [BIP 84 derivation scheme for P2WPKH based accounts](/using-wasabi/BI
 ### Public Key
 
 A public key is calculated by multiplying the private key to the generator point of an elliptic curve.
-Given only the public key, the private key cannot be calculated. 
+Given only the public key, the private key cannot be calculated.
 Anyone can encrypt a message using a public key.
 This cyphertext message can only be decrypted through the related private key.
 Given a public key and a signature over a message, anyone can verify that the signer had knowledge of the private key and message.
--- a/docs/glossary/Glossary.md
+++ b/docs/glossary/Glossary.md
@@ -104,6 +104,7 @@ Given only the public key, the private key cannot be calculated.
 Anyone can encrypt a message to the public key, which can only be decrypted with the knowledge of the private key.
 Given a public key and a signature over a message, anyone can verify that the signer had knowledge of the private key and message.
 In Bitcoin, the public key is the pseudonymous identity of the owner of a coin.
+Read more: [Bitcoin private keys](/using-wasabi/Receive.html#bitcoin-public-keys-and-addresses)
 :::
 
 :::details
--- a/docs/glossary/Glossary.md
+++ b/docs/glossary/Glossary.md
@@ -99,7 +99,7 @@ Read more: [BIP 84 derivation scheme for P2WPKH based accounts](/using-wasabi/BI
 :::details
 ### Public Key
 
-A public key is calculated by multiplying the private key to the generator point of an elliptical curve.
+A public key is calculated by multiplying the private key to the generator point of an elliptic curve.
 Given only the public key, the private key cannot be calculated. 
 Anyone can encrypt a message using a public key.
 This cyphertext message can only be decrypted through the related private key.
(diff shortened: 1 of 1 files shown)


docs/glossary/Glossary.md
```diff
@@ -96,6 +96,17 @@ The scriptPubKey is also modified.
 Read more: [BIP 84 derivation scheme for P2WPKH based accounts](/using-wasabi/BIPs.md#bip-84-derivation-scheme-for-p2wpkh-based-accounts)
 :::
 
+:::details
+### Public Key
+
+A public key is calculated by multiplying the private key to the generator point of an elliptical curve.
+Given only the public key, the private key cannot be calculated. 
+Anyone can encrypt a message to the public key, which can only be decrypted with the knowledge of the private key.
```
SimoneBovi: Anyone can encrypt a message using a public key. Then this message can only be decrypted through the related private key.

docs/glossary/Glossary.md
```diff
@@ -96,6 +96,17 @@ The scriptPubKey is also modified.
...
+Given a public key and a signature over a message, anyone can verify that the signer had knowledge of the private key and message.
+In Bitcoin, the public key is the pseudonymous identity of the owner of a coin.
```
SimoneBovi: Maybe I would specify the hash of the public key.

docs/glossary/Glossary.md
```suggestion
Anyone can encrypt a message using a public key. Then this message can only be decrypted through the related private key.
```

docs/glossary/Glossary.md
MaxHillebrand: well, you cannot say that in general... 
P2PK and taproot for example are not hashed, yet still the pseudonymous identity.

SimoneBovi: Yeah sorry, I meant that we could change it with something like "the public key (or its hash)", but if this is confusing, nevermind.

docs/glossary/Glossary.md
```diff
@@ -96,6 +96,18 @@ The scriptPubKey is also modified.
 Read more: [BIP 84 derivation scheme for P2WPKH based accounts](/using-wasabi/BIPs.md#bip-84-derivation-scheme-for-p2wpkh-based-accounts)
 :::
 
+:::details
+### Public Key
+
+A public key is calculated by multiplying the private key to the generator point of an elliptical curve.
```
```suggestion
A public key is calculated by multiplying the private key to the generator point of an elliptic curve.
```

docs/glossary/Glossary.md
```diff
@@ -96,6 +96,18 @@ The scriptPubKey is also modified.
...
+Given only the public key, the private key cannot be calculated.
```
```suggestion
Given only the public key, the private key cannot be calculated.
```

docs/glossary/Glossary.md
```diff
@@ -96,6 +96,18 @@ The scriptPubKey is also modified.
 Read more: [BIP 84 derivation scheme for P2WPKH based accounts](/using-wasabi/BIPs.md#bip-84-derivation-scheme-for-p2wpkh-based-accounts)
 :::
 
+:::details
+### Public Key
+
+A public key is calculated by multiplying the private key to the generator point of an elliptical curve.
+Given only the public key, the private key cannot be calculated. 
+Anyone can encrypt a message using a public key.
+This cyphertext message can only be decrypted through the related private key.
```
```suggestion
This encrypted message (cyphertext) can only be decrypted through the related private key.
```

docs/glossary/Glossary.md
```diff
@@ -96,6 +96,18 @@ The scriptPubKey is also modified.
...
+Given a public key and a signature over a message, anyone can verify that the signer had knowledge of the private key and message.
```
```suggestion
Given a public key and a signature over a message, anyone can verify that the signer had the private key that signed the message.
```

yahiheb:
> NACK.
> I think Bitcoin basics should not be in Wasabi glossary.

I think @RiccardoMasutti has a point here.

docs/glossary/Glossary.md
MaxHillebrand: what I mean here is that you need to have the actual knowledge of both the private key and the message. This is explicitly important in bitcoin, because the message is the spending tx transferring ownership. So even if I have the private key, but do not know the spending tx, then I cannot sign the ownership transfer. So I must explicitly agree to the tx.

MaxHillebrand:
This is a general question of if the basics should be included. I pinged you in Slack for a proper conversation.

But I think that two guidelines for if to add content make sens:
- the term is used elsewhere in the docs already
- the term is important to understand a concept which is explained elsewhere in the docs

According to this, public keys should most be included because we mention it many many times in the docs.

docs/glossary/Glossary.md
yahiheb: I updated my suggestion.

PR #436, opened 2019-12-22
[GLOSSARY] address

MaxHillebrand:
adds address to glossary


RiccardoMasutti:
NACK.

MaxHillebrand:
Address is mentioned countless times in the docs, and it is an important privacy topic, as it can be misused.

PR #438, opened 2019-12-22
[GLOSSARY] blockchain

MaxHillebrand:
adds timechain to glossary

### DIFF
--- a/docs/glossary/Glossary.md
+++ b/docs/glossary/Glossary.md
@@ -122,7 +122,7 @@ Read more: [Wasabi Wallet under the hood](/building-wasabi/TechnicalOverview.md#
 ### Timechain
 
 The timechain is the one chain of valid Bitcoin blocks with the most accumulated proof of work.
-It is an append-only data structure with new blocks being added on average every 10 minutes.
+It is an append-only data structure with new blocks being added on average every 10 minutes, thus some call it blockchain.
 Because of its uniquely verifiable and global consensus, it is a revolutionary standard for time.
 :::
 
--- a/docs/glossary/Glossary.md
+++ b/docs/glossary/Glossary.md
@@ -119,6 +119,14 @@ Read more: [Wasabi Wallet under the hood](/building-wasabi/TechnicalOverview.md#
 :::
 
 :::details
+### Timechain
+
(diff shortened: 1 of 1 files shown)


docs/glossary/Glossary.md
```diff
@@ -118,6 +118,14 @@ The method is used by some lightweight Bitcoin clients.
 Read more: [Wasabi Wallet under the hood](/building-wasabi/TechnicalOverview.md#wasabi-wallet-under-the-hood)
 :::
 
+:::details
+### Timechain
+
+The timechain is the one chain of valid Bitcoin blocks with the most accumulated proof of work.
+It is an append only data structure with new blocks being added on average every 10 minutes.
```
SimoneBovi: It is an append-only data structure with new blocks being added on average every 10 minutes.

```suggestion
It is an append-only data structure with new blocks being added on average every 10 minutes.
```

RiccardoMasutti:
NACK.

yahiheb:
Shouldn't we say it is synonym to `Blockchain`?

docs/glossary/Glossary.md
```diff
@@ -118,6 +118,14 @@ The method is used by some lightweight Bitcoin clients.
 Read more: [Wasabi Wallet under the hood](/building-wasabi/TechnicalOverview.md#wasabi-wallet-under-the-hood)
 :::
 
+:::details
+### Timechain
+
+The timechain is the one chain of valid Bitcoin blocks with the most accumulated proof of work.
+It is an append-only data structure with new blocks being added on average every 10 minutes.
```
MaxHillebrand:
```suggestion
It is an append-only data structure with new blocks being added on average every 10 minutes, thus some call it blockchain.
```

yahiheb: ACK. I think most call it blockchain.

SimoneBovi: I agree with @yahiheb. Related to that, I think it could be confusing to use the term timechain on the Doc homepage.

yahiheb:
Maybe we should replace `Timechain` by `Blockchain` and mention that some people call it `Timechain`.

MaxHillebrand:
`blockchain` is mentioned 15 + times in the docs, thus it should be included in the glossary.

PR #442, opened 2019-12-22
[GLOSSARY] blockchain analysis

MaxHillebrand:
adds blockchain analysis to glossary.


RiccardoMasutti:
NACK.
Should be added as `Taint` synonymous

MaxHillebrand:
> Should be added as Taint synonymous - @RiccardoMasutti 

But the two are very different terms...

PR #440, opened 2019-12-22
[GLOSSARY] difficulty target

MaxHillebrand:
adds difficulty target to glossary

PR #444, opened 2019-12-23
[STRUCTUR] separate glossary

MaxHillebrand:
This changes no content, but only separates the glossary into two files, `Bitcoin in General` and `Privacy and Wasabi` with a base `README` which links to the two. 

This is the same setup as for the FAQ.

It might give merging conflicts with the open glossary PRs, but I'll deal with them.


MaxHillebrand:
This is ready for conceptual review and possible merge. 
The only open PR that is not included in this restructure is #440, which probably will not be included since it's too off-topic.

yahiheb:
I tried to remove `Glossary.md` file but the checks haven't completed yet, if they don't we should revert my last commit and do as @MaxHillebrand suggested.

MaxHillebrand:
Yes @yahiheb, I tried the same and stumbled upon the same issue, the checks won't even start... 
So I manually reverted the removal, not sure if I can do it for your commit too?

MaxHillebrand:
after the merge, the `Glossary.md` was still there, so I deleted it manually in b359bfd

screenshot_20191225_214008

PR #445, opened 2019-12-24
add algolia search config

### LINKED ISSUE #38: [FEATURE] Algolia for Activate Content Search
Out-of-the-box, VuePress provides a search box that is looking through only the headers of all the markdown files. As far as I understand, there is an opt-in extension to allow the search for the text in the files.

I think we should activate this, but it might lead to a cluttered search experience.

dennisreimann:
Closes #38.

### DIFF
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -90,6 +90,10 @@ module.exports = {
     docsDir: "docs",
     editLinks: true,
     lastUpdated: 'Last Updated',
+    algolia: {
+      indexName: 'wasabiwallet',
+      apiKey: 'c9d9b7688e0f9e6d0ed534655321a424'
+    },
     nav: [
       {
         text: "Why Wasabi",


MaxHillebrand:
it would be nice to see more search results, for example here, when I first type in `ch` it shows only 3 entries for `change`
screenshot_20191224_095938
and when I then type in `change`, it shows 5.
screenshot_20191224_095947

It might be that there is a limit on 5 search results?

MaxHillebrand:
FYI, the search engine also finds typos and shows results for the fixed word.
For example here I type in `asd`, and it shows results for `and` & `add`.

screenshot_20191224_101019

MaxHillebrand:
And also FYI, the `Search by Algolia` logo is added because wasabi is a free & open source project and thus Algolia graciously provides their tool gratis, though with the caveat of this small advertisement. I think this is very fair.

yahiheb:
Great job guys.

yahiheb:
The problem I see now if you type `HOW DOES WASABI DOWNLOAD A RELEVANT BLOCK?` which is a FAQ you don't see it suggested.

MaxHillebrand:
I also thought that there was an issue with the faq search resultes [#450] but I think everything is working properly, it simply does not say that the result is part of the FAQ, but it is...

screenshot_20191224_161925

yahiheb:
I am not sure I like this. If you type `country` you get many suggestions which have nothing to do with what I am looking for like `contributions` `contributor` `control` ....

molnard:
@dennisreimann that is beautiful, congratulation!

PR #448, opened 2019-12-24
Increase number of search results

### LINKED ISSUE #446: Increase number of search results
### Problem

The new Algolia search engine #445 is very powerful. But it only shows 5 search results at any time.

When I search for `ch` it shows only 3 entries in `Change.md` and 2 others.
screenshot_20191224_095938

When I then search for `change`, it shows 5 entries.
screenshot_20191224_095947

### Solution

Show more than just 5 search results, though it might clutter the UI when the search box is too large...

dennisreimann:
As the search results box is not scrollable it does not make sense to increase `hitsPerPage` parameter to more than 25. Closes #446.

### DIFF
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -92,7 +92,11 @@ module.exports = {
     lastUpdated: 'Last Updated',
     algolia: {
       indexName: 'wasabiwallet',
-      apiKey: 'c9d9b7688e0f9e6d0ed534655321a424'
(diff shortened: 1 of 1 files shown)


MaxHillebrand:
Tested, the box is now huge, and it looks like it is going below my screen. As you say, it is not scrollable. 
screenshot_20191224_124642

Maybe we can add a right side search menu, similar to how the [algolia doc](https://www.algolia.com/doc/) has it?

PR #457, opened 2019-12-28
Add smooth scroll to documentation

RiccardoMasutti:
It gives a much better look to the website in general. The best is when an internal link (#) is clicked, e.g. in the FAQs


RiccardoMasutti:
Click your favourite submenu from the navbar and enjoy the magic :)

Issue #451, opened 2019-12-24
Open FAQ dropdown when searched


MaxHillebrand:
In the old search engine, if I clicked on a FAQ search result, then it jumped to that header in the file and also opened the dropdown.

With the Algolia search engine, it does only jump to the ### header, but it does not open the dropdown, which makes it a bit cumbersome to search the FAQ, as the result is not immediately clear.

Can Algolia also open the dropdown?

dennisreimann:
> Can Algolia also open the dropdown?

Yes it can 😁

Issue #449, opened 2019-12-24
Make search results scrollable 


MaxHillebrand:
Apparently there is a way to have a [scrollable search result list](https://www.algolia.com/doc/guides/building-search-ui/resources/ui-and-ux-patterns/in-depth/infinite-scroll/js/). 
It would be nice to have a way to see ALL the search results, not just the top 25 [#448].

Maybe it would be a nice solution to have a right side search menu, scrollable or with pages, similar to the [Agolia docs](https://www.algolia.com/doc/).

![screenshot_20191224_133010](https://user-images.githubusercontent.com/30683012/71413293-c8207c00-2651-11ea-89b8-6ac4a9bdb2f8.png)

yahiheb:
That can be good, but also imagine you get 100 results that is too much.

PR #458, opened 2019-12-29
Algolia updates

### LINKED ISSUE #450: Algolia does not return FAQ entrie
It seems that Algolia does not search through the FAQs, even though if it is a `###` header, as for example `dust treshold`.

screenshot_20191224_133728

Is this broken on their server side because of missing indexes, or is it a mistake in our repository?

dennisreimann:
These commits address the issues #449 and #450/#451:

The first one is fixed when merged: It makes the search the rightmost item in the navbar so that the results can be displayed as a sidebar on the right. This gives it more space. In addition to that the CSS changes introduced here make the list scrollable.

The latter issues might be solved by merging, which leads to Algolia reindexing the site. I could not fully verify this locally, but Algolia seems to link to the closest previous headline. The commit makes the FAQ summary a headline, which might fix #450. If so, this way we could restore the old behaviour of directly opening the FAQ item, fixing #451.

Needs to be deployed so that Algolia reindexes and we can test based on that.

### DIFF
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -50,7 +50,7 @@ module.exports = {
             title = ''
           }
           const slug = slugify(title)
-          return `<details id="${slug}"><summary><a href="#${slug}" aria-hidden="true" class="header-anchor">#</a> ${title}</summary>`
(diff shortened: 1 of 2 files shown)


dennisreimann:
Thanks for merging, I‘ll check the rest later today 🤘🏻

PR #477, opened 2020-01-13
Add Custodial and Non-Custodial to Glossary

RiccardoMasutti:


docs/glossary/Glossary-GeneralBitcoin.md
```diff
@@ -91,6 +91,16 @@ When several nodes, usually most nodes on the network, all have the same blocks
...
 :::
 
+:::details
+### Custodial (Wallet or Mixer)
+
+A custodial wallet is any wallet where the private keys of your coins are in the hands of a third party.
+In essence, they operate similarly to traditional banks.
+
+A custodial mixer needs you to send your bitcoins to a central part before mixing.
+This means that you risk losing your money or being stolen.
```
yahiheb: I am not sure if the current sentence is correct.
Maybe we should say `you risk .... being stolen from`
or `you risk .... being robbed`

MaxHillebrand:
```suggestion
This means that you are at risk of being stolen from.
```

docs/glossary/Glossary-GeneralBitcoin.md
```diff
@@ -230,6 +240,17 @@ The `nonce` in a Bitcoin block is a 32-bit (4-byte) field whose value is set so
 The rest of the fields may not be changed, as they have a defined meaning.
 :::
 
+:::details
+### Non-Custodial (Wallet or Mixer)
+
+A non-custodial wallet is any wallet where the private keys of your coins are in your hands.
+Non-custodial wallets give you full control over your bitcoins.
+
+A non-custodial mixer like Wasabi eliminates the risk of funds disappearing or being stolen.
```
MaxHillebrand:
```suggestion
A non-custodial mixer like Wasabi eliminates the risk that the coordinator can steal your bitcoin.
```

docs/glossary/Glossary-GeneralBitcoin.md
```diff
@@ -230,6 +240,17 @@ The `nonce` in a Bitcoin block is a 32-bit (4-byte) field whose value is set so
...
+Each of the signatures are created on the participants’ computers, so anyone trying to connect the signatures is unable to alter the transaction or redirect the funds.
```
MaxHillebrand:
```suggestion
Each participant verifies the proposed transaction, and signs it only if it is valid.
This signature is only valid for this transaction, so any attempt to change the transaction after the fact is futile.
```

PR #497, opened 2020-01-18
Remove Daemon/API section [merge upon release]

yahiheb:
Maybe this should be removed since [RPC PR](https://github.com/zkSNACKs/WalletWasabi/pull/1449) was merged.

PR #503, opened 2020-01-20
Remove standalone daemon chapter

yahiheb:
As @MaxHillebrand pointed out [here](https://github.com/zkSNACKs/WasabiDoc/pull/497#issuecomment-576024121) maybe this should be removed.


MaxHillebrand:
I spoke with @lontivero, the RPC does not replace the daemon. You can try running it on master where the RPC is already merged.

```
~/WalletWasabi/WalletWasabi.Gui$ git rev-parse HEAD
3ddd95a821018698f91842440c7898d194eb0fbd
~/WalletWasabi/WalletWasabi.Gui$ dotnet run mix --wallet:breaking --mixall --keepalive --loglevel:info
2020-01-20 22:27:01 INFO	Daemon (27)	Wasabi Daemon started (35fe02cb-533a-4926-a186-e4349eb446ca).
2020-01-20 22:27:02 INFO	Global (751)	Wallet loaded: breaking.
Password: 
2020-01-20 22:27:07 INFO	Daemon (73)	Correct password.
2020-01-20 22:27:07 INFO	Global (177)	Config is successfully initialized.
```

So, the proper way would be using the daemon with --keepalive, and then use the RPC server to send commands. Otherwise, you could start wasabi gui, and then use the rpc.

Thus, I'm closing this PR.

PR #514, opened 2020-01-22
[FAQ] delete address

MaxHillebrand:
This adds a FAQ on how to delete an already generated address, by manually deleting the pubkey from the wallet file.

@lontivero @nopara73, I've tested and this works, but I am not sure if this is actually good advice... What do you think?

### DIFF
--- a/docs/FAQ/FAQ-UseWasabi.md
+++ b/docs/FAQ/FAQ-UseWasabi.md
@@ -295,6 +295,25 @@ When you import the wallet file into a new Wasabi client, then it will use this 
 :::
 
 :::details
+### How do I delete a generated address?
+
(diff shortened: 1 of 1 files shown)


yahiheb:
What if there are multiple address with the same label?

nopara73:
I think asking the wrong question is fine, but you should reply with the right answer: you cannot delete an address.

yahiheb:
@MaxHillebrand Don't forget to delete the branch and also this branch https://github.com/zkSNACKs/WasabiDoc/tree/RPC

Also I think it is better to add branches on your repo and then open a PR
https://github.com/zkSNACKs/WalletWasabi/pull/2988#issuecomment-576848863

PR #517, opened 2020-01-24
fixing links in the FAQ README

raindogdance:
Fixed the links [here](https://github.com/zkSNACKs/WasabiDoc/tree/master/docs/FAQ), as they are currently leading to 404 pages.

### DIFF
--- a/docs/FAQ/README.md
+++ b/docs/FAQ/README.md
@@ -14,8 +14,8 @@ Further the questions for those who are interested in contributing to Wasabi.
 
 ### Chapters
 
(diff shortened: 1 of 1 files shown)


RiccardoMasutti:
NACK:

https://docs.wasabiwallet.io/FAQ/

raindogdance:
Ahh, so the links need to be this way they are now in order to work correctly on the actual website (even though they're throwing 404 for the github linking). I think I got it :) 

Closing the PR.

raindogdance:
Yup, makes sense, thanks @MaxHillebrand and @RiccardoMasutti :)

PR #521, opened 2020-01-24
Optimize images

dennisreimann:
Lossless compression, saves our readers ~35% of bandwidth. 🏎


dennisreimann:
In this case I used [ImageOptim](https://imageoptim.com/), but we could also add a script that can be run for images added. This should be done manually imho, because processing the images takes some time and running it is only necessary for newly added images.

MaxHillebrand:
Ok, thanks @dennisreimann, then let's remember to every now and then again do this manually. I guess it won't be too many new pictures added anyhow, and all of the existing ones are now processed already.

PR #524, opened 2020-01-27
[Using] WasabiSetupVM.md - Add a VirtualBox guide

davterra:
Add instructions for the use of Virtualbox

@MaxHillebrand 
Here is the VirtualBox add-on to your Qubes instructions, as we discussed.

Note: 
I wasn't sure how to add images in a way that they might transfer upon merge. 
I saw some other doc files with imgur links, so I used that.
Please let me know if I need to do something different with the image links.

Also, since the VirtualBox installation/VM creation process is longer than in using Qubes, my additon is lengthy. 
If you think it should be broken out as a separate file, I can close this and resubmit my additions with a new file name.


RiccardoMasutti:
NACK. Most of the content is duplicated from [Oracle documentation](https://docs.oracle.com/cd/E97728_01/E97727/html/) and [VirtualBox user manual](https://www.virtualbox.org/manual/UserManual.html).

Imho we should link to the original documents and let only Wasabi-on-VM-related installation inside docs.wasabiwallet.io

davterra:
> NACK. Most of the content is duplicated from [Oracle documentation](https://docs.oracle.com/cd/E97728_01/E97727/html/) and [VirtualBox user manual](https://www.virtualbox.org/manual/UserManual.html).
> 
> Imho we should link to the original documents and let only Wasabi-on-VM-related installation inside docs.wasabiwallet.io

I see your point, but my attempt was to extract and convey only the vital steps from the Oracle/VB documentation, which can be overwhelming if you are unfamiliar with VirtualBox.
I think external links to background info can indeed be useful and save space. On the other hand, they also invite the user to leave our site. 
I'd like to try to keep them here, once they get here.

PulpCattel:
I agree with @RiccardoMasutti  that it's probably better to do the same as the Qubes guide, where there's a link to the relative documentation first and then only the Wasabi related stuff.

I'm thinking, as a side note, that the next addition could be a Whonix guide, which is arguably one of the best, if not the best, OS to use with Wasabi inside a VM or Qubes.

davterra:
> I agree with @RiccardoMasutti that it's probably better to do the same as the Qubes guide, where there's a link to the relative documentation first and then only the Wasabi related stuff.
> 
> I'm thinking, as a side note, that the next addition could be a Whonix guide, which is arguably one of the best, if not the best, OS to use with Wasabi inside a VM or Qubes.

I like that idea. It makes more sense to me than running Wasabi in Tails w/persistence, which we already have a guide for.
I would be happy to take that on. I've used Whonix for several years, both in Qubes and via VirtualBox.

PulpCattel:
> It makes more sense to me than running Wasabi in Tails w/persistence, which we already have a guide for.

Seems to me that they are two very different things with very different trade offs.
Using Whonix in a VM you are keeping your private/public keys on your host and the host could be used for various stuff. If your host is compromise so is your wallet/Whonix/VM. 
Tails protects you from that; the private/public keys are "there" only when Tails is running and you use it for one single purpose. 
Main benefit of Whonix is network level privacy, which is already solved by Wasabi without further help in most cases and in general is not the most pressing issue with Bitcoin base layer (this could change with Lightning though).

But yeah, Whonix is dope too and every user can pick his own trade offs.

davterra:
> > It makes more sense to me than running Wasabi in Tails w/persistence, which we already have a guide for.
> 
> Seems to me that they are two very different things with very different trade offs.
> Using Whonix in a VM you are keeping your private/public keys on your host and the host could be used for various stuff. If your host is compromise so is your wallet/Whonix/VM.
> Tails protects you from that; the private/public keys are "there" only when Tails is running and you use it for one single purpose.
> Main benefit of Whonix is network level privacy, which is already solved by Wasabi without further help in most cases and in general is not the most pressing issue with Bitcoin base layer (this could change with Lightning though).
> 
> But yeah, Whonix is dope too and every user can pick his own trade offs.

If I used Wasabi in Whonix, I would use a cloned workstation "Wasabi Wallet Only": to be used only when needed, and use a different workstation(s) for other needs.
That's one of the awesome things about VMs. The Whonix "gateway" VM will work with any "workstation" clone.
I guess my point is, that if you're going to forego the Read-Only feature of Tails, and install persistence, you might as well use Whonix.

PulpCattel:
@davterra 
> if I used Wasabi in Whonix, I would use a cloned workstation "Wasabi Wallet Only": to be used only when needed, and use a different workstation(s) for other needs.
That's one of the awesome things about VMs. The Whonix "gateway" VM will work with any "workstation" clone.

Unfortunately, this doesn't change anything. Your host is likely an unsafe machine with programs, software updates and so on. Even if a user is techy enough to use VM only and he keeps the host as clean as possible, your sensible information are still there. Even if you encrypt the VM or do other fancy stuff, the data are there. Any possible attack that compromises your host (and VM are not bulletproof by any mean), even if coming from something that has nothing to do with your "Wasabi only" VM, can compromise everything. 
Every time you do any action on this machine, you should be careful about the consequences of this action in regards to your sensible information. Tails fixes this.

> I guess my point is, that if you're going to forego the Read-Only feature of Tails, and install persistence, you might as well use Whonix.

This is incorrect. Persistance only means that you have an encrypted partition on your Tails USB/SD, you can launch Tails without decrypting the Persistance. 
It's not the same thing as Whonix, there's no host that keep running when you are not using the "Wasabi only" VM, there are no sensible information at stake when Tails is shut down. This is huge difference. 
The comparison with Whonix would be using an entire OS just for Whonix, where you have exclusively the "Wasabi only" VM and you do no other stuff a part from using the wallet in this OS. I'd argue that Tails is still better (I don't like trusting VirtualBox for example), but we could find similarities between this two setup. Here though, I would argue the opposite: if you have to dedicate an entire OS just for that, you might as well using Tails, which is portable, simpler, exc...

davterra:
> @davterra
> 
> > if I used Wasabi in Whonix, I would use a cloned workstation "Wasabi Wallet Only": to be used only when needed, and use a different workstation(s) for other needs.
> > That's one of the awesome things about VMs. The Whonix "gateway" VM will work with any "workstation" clone.
> 
> Unfortunately, this doesn't change anything. Your host is likely an unsafe OS with programs, software updates and so on. Even if a user is techy enough to use VM only and he keeps the host as clean as possible, your sensible information are still there. Even if you encrypt the VM or do other fancy stuff, the data are there. Any possible attack that compromises your host (and VM are not bulletproof by any mean), even if coming from something that has nothing to do with your "Wasabi only" VM, can compromise everything.
> Every time you do any action on this machine, you should be careful about the consequences of this action in regards to your sensible information. Tails fixes this.
> 
> > I guess my point is, that if you're going to forego the Read-Only feature of Tails, and install persistence, you might as well use Whonix.
> 
> This is incorrect. Persistance only means that you have an encrypted partition on your Tails USB/SD, you can launch Tails without decrypting the Persistance.
> It's not the same thing as Whonix, there's no host that keep running when you are not using the "Wasabi only" VM, there are no sensible information at stake when Tails is shut down. This is huge difference.
> The comparison with Whonix would be using an entire OS just for Whonix, where you have exclusively the "Wasabi only" VM and you do no other stuff a part from using the wallet in this OS. I'd argue that Tails is still better (I don't like trusting VirtualBox for example), but we could find similarities between this two setup. Here though, I would argue the opposite: if you have to dedicate an entire OS just for that, you might as well using Tails, which is portable, simpler, exc...

That's a great perspective that, as someone who hasn't really used Tails, I need to consider.
Thanks for taking the time to convey all of that

davterra:
> This PR is a great example on how much I like out collaboration on this project!
> 
> Thanks @davterra for writing this very thorough guide, @RiccardoMasutti and @PulpCattel for reviewing and providing very insightful feedback!
> 
> A lot has been said already, and I agree with most. I think this is a rather long and thorough explanation, and it would be better if it is more brief with a link to the well documented VirtualBox guide. Yes, this will send users out of the Wasabi docs, but I believe in this case it is alright. Some of this intro text on why virtualization is awesome can be put at the very first header, even before the Qubes header, as it is a very good explanation.
> 
> The step by step guide on how to setup Wasabi is also very good, that can be reviewed and then merged. I don't think it needs a separate chapter, it is good here together with Qubes.
> 
> Regarding Images, they can be added in the `docs/.vuepress/public` folder, and then referenced in the md file with ``, as [described here](https://github.com/zkSNACKs/WasabiDoc/#embedding-images). The file upload cannot be done in the GitHub Gui, I believe, so do this in terminal.
> 
> A whonix guide would be dope as well, thanks Dav for writing that! but that's off topic in context of this PR..
Thanks for your thoughts @MaxHillebrand 
Do you want me to close this PR and resubmit a condensed version per your recommendations, or just prune the content in this PR?

And yeah, I'd be happy to write up a whonix version, but I didnt mean to imply that I would add it to this PR. This one has enough to chew over.

Thanks to @RiccardoMasutti and @PulpCattel for your feedback! I think I knew all along that you were correct in your assessment of making it shorter and more Wasabi-relevant.

PR #530, opened 2020-01-29
Corrections regarding Ledger Live

dasilvarosa:
1. Ledger Live does not send xpubs to any server.
2. Ledger does not store public addresses or IP addresses to deanonymize users. 
3. Since 2. cannot be proven, it is indeed recommended to run a node for better privacy.
4. Removed the part recommending to generate a new seed after setting up a Ledger hardware wallet because the xpub is never sent to Ledger's server. An xpub is therefore never _compromised_ in this context. 
5. Reworded the warning to keep using Ledger Live for firmware updates.

### DIFF
--- a/docs/FAQ/FAQ-UseWasabi.md
+++ b/docs/FAQ/FAQ-UseWasabi.md
@@ -1025,7 +1025,8 @@ All of them!
 This is thanks to awesome Hardware Wallet Interface of Bitcoin core.
 
 However, not all of the many hardware wallets have been tested and reviewed by the developers.
-Everything seems to work fine with the ColdCard, BitBox, Trezor, Ledger and KeepKey.
+Everything seems to work fine with the ColdCard, BitBox, Trezor, 
+and KeepKey.
 ::::
 
 :::details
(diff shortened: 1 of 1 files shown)


docs/FAQ/FAQ-UseWasabi.md
```diff
@@ -1097,21 +1097,13 @@ You can't do that directly, you have to send the bitcoins (in small portions > 0
...
-With this information they know exactly how many bitcoins you have and in what transactions you spend them.
-To avoid that you can use a Ledger hardware in combination with Wasabi as software interface, and because Wasabi does not leak your xpub, your transaction history is not shared with anyone.
-Make sure you are generating a completely new wallet with a new seed phrase backup so that the compromised old wallet is no longer an issue.
-You can also go into the Settings of your Ledger and create a PIN-specific passphrase.
-
-To create a new Ledger Wallet *offline* and make sure that your newly created wallet doesn't leak any information to Ledger's servers you must not use Ledger Live (neither to configure it for the first time, nor to use it).
-After that, you can use the wallet in combination with Wasabi without having concerns about your privacy!
+No, when using the Ledger Live application to manage your Ledger device and coins, your extended public key is never sent to Ledger's servers. Ledger Live stores your xpub locally on your computer or smartphone and encrypts it with your optional password.
```
MaxHillebrand:
```suggestion
No, when using the Ledger Live application to manage your Ledger device and coins, your extended public key is never sent to Ledger's servers.
Ledger Live stores your xpub locally on your computer or smartphone and encrypts it with your optional password.
```

docs/FAQ/FAQ-UseWasabi.md
```diff
@@ -1097,21 +1097,13 @@ You can't do that directly, you have to send the bitcoins (in small portions > 0
...
+Ledger Live sends individual public addresses to Ledger's blockchain explorers to calculate the balance of your accounts, though these are not stored during normal operation. For increased privacy, users should run their own node or may use their Ledger hardware wallet in combination with Wasabi as a software interface.
```
MaxHillebrand:
```suggestion
Ledger Live sends individual public addresses to Ledger's blockchain explorers to calculate the balance of your accounts, though these are not stored during normal operation.
For increased privacy, users should run their own node or may use their Ledger hardware wallet in combination with Wasabi as a software interface.
```

MaxHillebrand:
```suggestion
Ledger Live sends individual public addresses to Ledger's blockchain explorers to calculate the balance of your accounts, though these are not stored during normal operation. For any privacy, users should run their own node or may use their Ledger hardware wallet in combination with Wasabi as a software interface.
```

sending individual addresses compromises privacy to a similar degree as sending the xpub.

dasilvarosa: Sending an individual address compromises the privacy of the individual address. Sending an xpub compromises all future addresses. I fail to see how privacy is harmed to a similar degree in these two cases.

MaxHillebrand: How many addresses do you send? I would assume it's a gap limit of 100 or so? In that sense, all addresses for the foreseeable future are doxed, so only after these gap limit addresses are spend does the user have a better privacy compared to xpub leak.

dasilvarosa: The address gap limit is 20 as prescribed in BIP44. It's safe to assume that this is the case for any client wallet using an external node, and not unique to Ledger Live. 

Still, **20 addresses** is not the same as **any possible future address** which could potentially be millions of addresses. For all practical purposes though, it's great to be transparent with users, but let's then generalize this to any non-validating wallet software and not just Ledger Live.

MaxHillebrand:
This should be closed in favor of #566 with a more clear explanation of the privacy leak.

PR #550, opened 2020-02-13
Indicate that all sending transactions should be labeled

yahiheb:
https://github.com/zkSNACKs/WalletWasabi/issues/3070#issuecomment-585175011
https://github.com/zkSNACKs/WalletWasabi/issues/3070#issuecomment-585611590

### DIFF
--- a/docs/using-wasabi/Send.md
+++ b/docs/using-wasabi/Send.md
@@ -61,7 +61,7 @@ Wasabi will calculate the checksum and notify you if the provided address is wro
 
 Similar to the `Receive` tab, you must label every new address with the observers who know that this is your address.
 In the `Send` tab an address is automatically generated to receive the change amount.
-A label is not required when you send an entire coin, as there will be no change left.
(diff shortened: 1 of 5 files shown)


docs/using-wasabi/Send.md
```diff
@@ -61,7 +61,7 @@ Wasabi will calculate the checksum and notify you if the provided address is wro
 
 Similar to the `Receive` tab, you must label every new address with the observers who know that this is your address.
 In the `Send` tab an address is automatically generated to receive the change amount.
-A label is not required when you send an entire coin, as there will be no change left.
+A label is not required (except in edge cases) when you send an entire coin, as there will be no change left.
```
MaxHillebrand: There is talk about adding the label mandatory, @lontivero is this going to be implemented? If yes, then this should be removed.

```suggestion
```

yahiheb: I am not sure if it will be implemented or not, I say we merge this and it is implemented we remove that.

PR #551, opened 2020-02-13
Add build step

yahiheb:

PR #555, opened 2020-02-13
Remove unnecessary build step

yahiheb:
https://github.com/zkSNACKs/WasabiDoc/pull/551#issuecomment-585838102

### DIFF
--- a/docs/FAQ/FAQ-Installation.md
+++ b/docs/FAQ/FAQ-Installation.md
@@ -171,10 +171,10 @@ Yet in the meantime there are many commits to the latest master branch, not just
 If you cannot wait until the next release, and you want to experience the most cutting-edge version of Wasabi, then you can [build the source code](/using-wasabi/BuildSource.md).
 
 The only two required tools are [Git](https://git-scm.com/downloads) and [.NET Core ${dotnetVersion} SDK](https://www.microsoft.com/net/download) for "Building Apps".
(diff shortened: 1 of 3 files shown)


MaxHillebrand:
Wait, I thought you said that... `Whenever you pull new commits you need to build.`
Does this not include the update or dev branch? so it is only upon the first time wasabi is compiled?

yahiheb:
Whenever you pull new commits (changes) you have to build, so when you next run you will see those new changes.
To `update-wasabi` or to `checkout a developer branch` you don't need the `dotnet build` step, but of course if you want to run Wasabi after checking out a developer branch you definitely need to build.

MaxHillebrand:
I'm confused...

Why would I download a dev branch, if not with the intent to run the software? If every time when I download a branch and want to run it, I must `dotnet build`, then why not add it to the docs?

yahiheb:
> If every time when I download a branch and want to run it, I must dotnet build, then why not add it to the docs?

There is nothing wrong in adding it to the docs.
If we do then we should add `dotnet run` after each `dotnet build` because why would you build if you don't run the software.

MaxHillebrand:
I just spoke with Adam about it.
Wasabi's `dotnet run` includes the building of the source code. So in most cases, when everything works properly, you should never need to run `dotnet build`. However, there were some edge case bugs on some OS where you would need to use `dotnet build` first. 

In an effort to de-clutter, I would say we remove `dotnet build` from the docs entirely, and just do `dotnet run`.

MaxHillebrand:
@yahiheb can you please expand this PR to remove all the `dotnet build`?

Issue #70, opened 2019-08-05
How to format images?


MaxHillebrand:
How can I format the position of a picture?
For example, how can it be nested on the side of a paragraph with breaking lines at the boarders?

dennisreimann:
Do you have any examples for images you‘d like to format specifically?

MaxHillebrand:
For example in the installation instruction, there is a picture of the new website download icon, this is a rather small image, and I think it would be nice to nest it on the right / left side of a paragraph with wrapped lines.

MaxHillebrand:
we got all the images in `/.vuepress/public/` in `.png` format.

PR #559, opened 2020-02-13
README: How To Nest Images and Text

davterra:
This insert to the README will show how to nest and format images and text side-by-side in the docs.

As the README.md is seemingly hosted on github, I left the images referenced in my code as absolute links, rather than the "in-docs" relative links normally used for images. 
Seeing as how the Wasabika images on the bottom of the page are absolute links, I'm assuming this is the way to do it for this page.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ They can be embedded via the following markdown tags:
 
 ### Nesting Images Alongside Text
 
-Images may be nested on one side or the other of a block of text, and the image size may be customized, 
+Images may be nested on one side or the other of a block of text, and the image size may be customized, using html tags.
 using html tags.
 
 As shown in the following examples, you can:
--- a/README.md
+++ b/README.md
@@ -93,7 +93,6 @@ They can be embedded via the following markdown tags:
 ### Nesting Images Alongside Text
 
 Images may be nested on one side or the other of a block of text, and the image size may be customized, using html tags.
-using html tags.
 
 As shown in the following examples, you can:
 
--- a/README.md
+++ b/README.md
@@ -90,6 +90,41 @@ They can be embedded via the following markdown tags:
 ![](/IMAGENAME.png)
 ```
 
+### Nesting Images Alongside Text
+
+Images may be nested on one side or the other of a block of text, and the image size may be customized, 
+using html tags.
+
+As shown in the following examples, you can:
+
+- Use `float` tags to place the image on the left or right side of the text.
+- Use `width` and `height` tags to size the image as desired.
+- Use `margin-left`, `margin-right`, `margin-top`, or `margin-bottom` tags to place a buffer space on any side of the image.
+- Use `<p>` and `</p>` paragraph tags to contain the text.
+- Use `<br>` line-breaks to format and position text placement relative to the images.
+
+#### HTML tag example with the image on the left:
+
+`<img src="https://docs.wasabiwallet.io/Logo_with_text.png"
+style="float:left;width:250px;height:250px;margin-right:50px;margin-left:50px;margin-top:0px;margin-bottom:0px"><p><br>This code will result in an image nested on the left side of a paragraph, and sized at 250 x 250 pixels, with a 50 pixel margin on the left and right side of the image, and the text centered vertically, relative to the image.</p>`
+
+__Result:__
+
+<img src="https://docs.wasabiwallet.io/Logo_with_text.png"
+style="float:left;width:250px;height:250px;margin-right:50px;margin-left:50px;margin-top:0px;margin-bottom:0px"><p><br>This code will result in an image nested on the left side of a paragraph, and sized at 250 x 250 pixels, with a 50 pixel margin on the left and right side of the image, and the text centered vertically, relative to the image.</p>
+<br>
+
+#### HTML tag example with the image on the right:
+
+`<img src="https://docs.wasabiwallet.io/Logo_with_text.png"
+style="float:right;width:400px;height:300px;margin-right:20px;margin-left:10px;margin-top:0px;margin-bottom:0px"><p>This code will result in an image nested on the left side of a paragraph, and sized at 400 x 300 pixels, with a 20 pixel margin on the right, and a 10 pixel margin on left side of the image, and the text *not* centered vertically, relative to the image.</p>`
+
+__Result:__
+
+<img src="https://docs.wasabiwallet.io/Logo_with_text.png"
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -90,6 +90,41 @@ They can be embedded via the following markdown tags:
 ![](/IMAGENAME.png)
 ```
 
+### Nesting Images Alongside Text
+
+Images may be nested on one side or the other of a block of text, and the image size may be customized,
```
yahiheb:
```suggestion
Images may be nested on one side or the other of a block of text, and the image size may be customized, using html tags.
```

README.md
```diff
@@ -90,6 +90,41 @@ They can be embedded via the following markdown tags:
 ![](/IMAGENAME.png)
 ```
 
+### Nesting Images Alongside Text
+
+Images may be nested on one side or the other of a block of text, and the image size may be customized, 
+using html tags.
```
yahiheb:
```suggestion
```

yahiheb: Don't forget to commit this as well.

MaxHillebrand:
@davterra still the README shows two gigantic images below the text...

davterra:
> ACK, thanks for the detailed write up!!
> 
> Though, it seems that your example is not working properly in the github README?
> screenshot_20200214_133419

Hmmm.....I don't know what to do about that. I tested it using the npm server with both relative and absolute links and it worked both ways. I could only hope and assume that it would work on github.
Maybe they have stricter markdown rules about inline html. 

If github doesnt work, I can rename this file and PR and we could put it in the regular docs to be placed with CodingConventions.

davterra:
> > ACK, thanks for the detailed write up!!
> > Though, it seems that your example is not working properly in the github README?
> > screenshot_20200214_133419
> 
> Hmmm.....I don't know what to do about that. I tested it using the npm server with both relative and absolute links and it worked both ways. I could only hope and assume that it would work on github.
> Maybe they have stricter markdown rules about inline html.
> 
> If github doesnt work, I can rename this file and PR and we could put it in the regular docs to be placed with CodingConventions.

More specifically, I meant to add the entire README page along with my addition to the docs...not just my addition. Call it DocumentationConventions.md

MaxHillebrand:
Well @davterra, if it compiles properly on vuepress, but wrongly on github, then I say in the README, only put the code block example, and not a live rendered example, or maybe a screenshot to the compiled vuepress.

README.md
```diff
@@ -90,6 +90,40 @@ They can be embedded via the following markdown tags:
...
 ```
 
+### Nesting Images Alongside Text
+
+Images may be nested on one side or the other of a block of text, and the image size may be customized, using html tags.
+
+As shown in the following examples, you can:
+
+- Use `float` tags to place the image on the left or right side of the text.
+- Use `width` and `height` tags to size the image as desired.
```
dennisreimann: Specifying `height` should not be necessary, as just using with will suffice and the image will keep the aspect ration. The `width` should be a viewport relative unit (e.g. `%` or `vw`) as this will work best in all viewport sizes.

Having said that it'd be a better option to specify classes to deal with different situations/images, as it might be best to have images displayed full width and not besides text on mobile.

README.md
```diff
@@ -90,6 +90,40 @@ They can be embedded via the following markdown tags:
...
+- Use `margin-left`, `margin-right`, `margin-top`, or `margin-bottom` tags to place a buffer space on any side of the image.
```
dennisreimann: This could also be obsoleted with the classes approach talked about above. We'd also ensure that we have consistent spacing and formatting everywhere and not have them vary across the images used in the docs.

README.md
```diff
@@ -90,6 +90,40 @@ They can be embedded via the following markdown tags:
...
+- Use `<p>` and `</p>` paragraph tags to contain the text.
+- Use `<br>` line-breaks to format and position text placement relative to the images.
```
dennisreimann: The usage of `<br>` in the example isn't what I'd expect. In special cases it might work to have text and images centered, but that depends on the viewport width. 
Also the paragraph would need a following `<div style="clear:both"></div>` to clear the float.

README.md
```diff
@@ -90,6 +90,40 @@ They can be embedded via the following markdown tags:
...
+
+#### HTML tag example with the image on the left:
+
+`<img src="/Logo_with_text.png"
```
dennisreimann: This should be embedded in a code fenced html block: 
````
```html
<img src="/Logo_with_text.png" ...
```

davterra:
Thanks for your very helpful thoughts @dennisreimann . Using a % instead of an absolute size should have been obvious to me. I haven't built a website from scratch in several years, and I forget stuff rather quickly, it seems.....

I'm going to give this another try, I guess by adding a couple of basic classes in index.styl
e.g., ".image-align-left" and ".image-align-right" that essentially just split the image/text space in half,
left & right sides.

I'll also work on some @media setting(s) to allow for smaller mobile displays, as that is indeed important and I had forgotten about it. 

Having said all that, I'm not sure if creating just  a couple of base-classes will suffice as a "one-size-fits-all" solution, but going beyond that doesnt seem worth the effort to me.

@MaxHillebrand this effort is in response to your issue #70, (Although I'd like to have the image-nesting ability, also) so since you are the most prolific contributor to the docs, I welcome any thoughts you have on how this would best work for you.

dennisreimann:
An alternative/extension would be to have separate containers like `:::image-left` for that. There's a bit of markup involved to make alignment work properly (like clearing the floats) and that could abstract the complexity away (or at least shift it toward something more intention revealing).

MaxHillebrand:
Great insights @dennisreimann & @davterra ! 
I really have no clue about the best approach for this, so I leave it up to you both to figure out the right path forward. :+1:

yahiheb:
Is this really necessary to add guys?
We already have https://github.com/zkSNACKs/WasabiDoc#embedding-images & https://github.com/zkSNACKs/WasabiDoc#embedding-shields

davterra:
> Is this really necessary to add guys?
> We already have https://github.com/zkSNACKs/WasabiDoc#embedding-images & https://github.com/zkSNACKs/WasabiDoc#embedding-shields

@yahiheb That became my concern and was the basis for my previous comment on this PR.

When I thought that this was an easy solution to an issue that Max had raised, and submitted the PR, I had overlooked the need to have it adaptable to various display sizes.

Adding that vital requirement has turned out to be more trouble than it is worth, IMO.

I think I'm going to close this PR and put it on my back-burner as a low-priority project to tinker with in my spare time, if for no other reason than to learn something new. 
I like @dennisreimann 's idea of creating separate containers.

If I stumble upon a solution that is easy and "no-fuss" for those who use it, I'll submit it in a new PR.

PR #572, opened 2020-02-17
Edit input registration time (one hour)

yahiheb:


yahiheb:
Maybe that is not necessary, I don't think we will keep changing the registration time so often.

PR #575, opened 2020-02-17
Make sub-titles clickable

yahiheb:
Similar to other sub-titles in that section.


MaxHillebrand:
Yes, this was in the original file too, before I moved it. 
However, it is the only file when we use this way `_**` to make headers bold, so I decided to keep it consistent with the rest of the docs, by making it `####`.

So, the current state is inconsistent, both for the docs in general and for the file specifically. We need to fix that. 

But I would favor to fix it to be consistent with the rest of the docs, by changing all in this file to `####`.

What do you say @yahiheb?

yahiheb:
> What do you say @yahiheb?

Done.

PR #577, opened 2020-02-17
[DECLUTTER BUILDING] update todo

MaxHillebrand:
This is ready for feedback update to the todo list. Though I would like to do a more complete refactor of this chapter. I think it is useful, but I think there is plenty of room for improvement. Does anyone have ideas on making this better?

### DIFF
--- a/docs/building-wasabi/ToDo.md
+++ b/docs/building-wasabi/ToDo.md
@@ -18,6 +18,7 @@ When you are ever bored and want to be productive, why not pick up one of these 
 
 - [Manually change the anonset shield](https://github.com/zkSNACKs/WalletWasabi/issues/1980)
 - RAM usage improvements
+- [Proper bitcoind integration](https://github.com/zkSNACKs/WalletWasabi/issues/2107)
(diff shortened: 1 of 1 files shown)


docs/building-wasabi/ToDo.md
```diff
@@ -27,7 +28,8 @@ When you are ever bored and want to be productive, why not pick up one of these
 ## Long Term Roadmap
 
 - [Rework the Wasabi UX](https://github.com/zkSNACKs/WalletWasabi/issues/1369)
-- [Proper bitcoind integration](https://github.com/zkSNACKs/WalletWasabi/issues/2107)
+- [Researching new CoinJoin mechanism](https://github.com/zkSNACKs/WasabiResearchClub/)
+- Complete rework of CoinJoin mechanism
```
yahiheb: Why is that?

MaxHillebrand: well, that is what the research is about, maybe something like zerolink + knappsack, or coinshuffle ++ etc. but it is redundant with the research club...

```suggestion
```

PR #580, opened 2020-02-18
Update and improve RPC.md

PulpCattel:
This includes various fixes and add the new [`createwallet`](https://github.com/zkSNACKs/WalletWasabi/pull/3050) method. 

A couple of things that maybe (maybe not and someone can quickly enlight me here) deserve an open issue:

*  `listkeys` method allows to query certain keys, however the provided examples `All already used keys` and `All unused locked keys (reserved for coinjoins)` use the exact same query but they are fairly different type of keys.

*  The `enqueue` method should require password but no parameter for it is specified as far as I can see. This generates a `Invalid password` response every time. 
A way to avoid that is to manually enqueue the coin providing the password through the GUI and then the command will work without password. This shouldn't be necessary and the `enqueue` method should allow for password.

Also a couple of explanations for some maybe strange changes:

* line 49 orders the methods following the correct page order.
* line 283 shows that adding a second label happens inside the same quotes and it's not a normal list (doing like this `["Daniel", "Alice"]` will return only "Daniel"). 
* line 292 shows that here the second label is shown as a normal list.

### DIFF
--- a/docs/using-wasabi/RPC.md
+++ b/docs/using-wasabi/RPC.md
@@ -126,7 +126,7 @@ curl -s --data-binary '{"jsonrpc":"2.0","id":"1","method":"getstatus"}' http://1
 
 ## createwallet
 
-Returns the twelve words of the freshly generated wallet. 
+Returns the twelve recovery words of the freshly generated wallet. 
 
 ```bash
 curl -s --data-binary '{"jsonrpc":"2.0","id":"1","method":"createwallet","params":["WalletName", "Password"]}' http://127.0.0.1:37128/ | jq                                                   
--- a/docs/using-wasabi/RPC.md
+++ b/docs/using-wasabi/RPC.md
@@ -46,7 +46,7 @@ Now you can use the following RPC commands to interact with your wallet, instead
 
 # Available methods
 
-The current version handles the following methods: `listunspentcoins`, `getstatus`, `getwalletinfo`, `getnewaddress`, `enqueue`, `dequeue`, `send`, `listkeys` and `stop`.
+The current version handles the following methods: `getstatus`, `createwallet`, `listunspentcoins`, `getwalletinfo`, `getnewaddress`, `send`, `gethistory`, `listkeys`, `enqueue`, `dequeue`, and `stop`.
 They can be used as follows:
 
 ## getstatus
@@ -124,6 +124,61 @@ curl -s --data-binary '{"jsonrpc":"2.0","id":"1","method":"getstatus"}' http://1
 }
 ```
 
+## createwallet
+
+Returns the twelve words of the freshly generated wallet. 
+
+```bash
+curl -s --data-binary '{"jsonrpc":"2.0","id":"1","method":"createwallet","params":["WalletName", "Password"]}' http://127.0.0.1:37128/ | jq                                                   
+```
+
+```json
+{                                                                                                            
+  "jsonrpc": "2.0",
+  "result": "jazz garment survey smart cricket child pizza reform physical alien envelope lesson",
+  "id": "1"
+}
+```
+
+In case we try to generate a wallet with an already existing name it will return:
+
+```json
+{                                                                                                            
+  "jsonrpc": "2.0",
+  "error": {
+    "code": -32603,
+    "message": "Wallet name is already taken."
+  },
+  "id": "1"
+}
+```
+
+In case we try to generate a wallet with a reserved wallet name it will return:
+
+```json
+{                                                                                                            
+  "jsonrpc": "2.0",
+  "error": {
+    "code": -32603,
+    "message": "Invalid wallet name."
+  },
+  "id": "1"
+}
+```
+
+In case we try to generate a wallet with too long of a password it will return:
+
+```json
+{
+  "jsonrpc": "2.0",
+  "error": {                                                                                                  
+    "code": -32603,
+    "message": "Password is too long (Max 150 characters)."
+  },
+  "id": "1"
+}
+```
+
 ## listunspentcoins
 
 Returns the list of confirmed and unconfirmed coins that are unspent.
@@ -172,6 +227,7 @@ curl -s --data-binary '{"jsonrpc":"2.0","id":"1","method":"listunspentcoins"}' h
 ```
 
 In case there is no wallet open it will return:
+
 ```json
 {
   "jsonrpc": "2.0",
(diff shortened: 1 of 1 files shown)


docs/using-wasabi/RPC.md
```diff
@@ -124,6 +124,61 @@ curl -s --data-binary '{"jsonrpc":"2.0","id":"1","method":"getstatus"}' http://1
 }
 ```
 
+## createwallet
+
+Returns the twelve words of the freshly generated wallet.
```
yahiheb:
```suggestion
Returns the twelve recovery words of the freshly generated wallet. 
```

lontivero:
I can't review this but i will let here a few comments:

* There's only one method 'listkeys' which returns all the keys. The rest is linux voodoo that should not be part of the documentation.
* `enqueue` should accept a password but given we don't provide tls i decided to do not pass that through an insecure channel. However, after that someone requested to add the password to the `send` method and i did it. At last, the password was also required to create a wallet and i added it there too so, there is an inconsistency now. I should add the password and implement the communication over tls.
* The label parameter accepts a string, not an array. That is clear in the documentation. If you want to label a coin as Daniel and Alice you have to do it as you would do it in the gui: "Daniel, Alice"

PulpCattel:
Thanks @lontivero.

@MaxHillebrand this should be a rather small PR and the biggest change should be the new added method. All the rest is either fixing or clarifying. Aniway, I'm happy to wait for a review from the dev team. 
In the meantime, let's try to summarize it starting from the lontivero's comments:

* we should remove the extra query examples, cause they are Linux related and not Wasabi RPC related, and let only the one `listkeys` example that returns all the keys. This should be done in this PR imho. 
* `Enqueue` method should allow for password and this is something to work on. There are no changes relative to this in this PR, so I say keep it like this for now and change it eventually in a new PR when the `enqueue` method will be completed. 
* The `getnewaddress` method requires label and the example request for it looked like this `"params":["Daniel"]` before that PR.
Since there are square brackets, intuitively I would do like this to add a second label: `["Daniel", "Alice"]`. But, as lontivero confirmed, the parameter for the label is a string and not a list (AKA array) and so the correct form for it is `["Daniel, Alice"]`. At the same time though, in the result, the label will be a list `["Alice", "Daniel"]`.
I basically just added a second label in the example to make this clear, but maybe I'm the only one to find this useful and we can revert to `["Daniel"]`.

The other changes should be more easy to follow, but please tell me what's unclear.

docs/using-wasabi/RPC.md
```diff
@@ -232,7 +289,7 @@ curl -s --data-binary '{"jsonrpc":"2.0","id":"1","method":"getnewaddress","param
   "result": {
     "address": "tb1qdskc4y529ayqkqrddknnhdjqwnqc9wzl8940pn",
     "keyPath": "84'/0'/0'/0/30",
-    "label": "Daniel",
+    "label": ["Alice", "Daniel"],
```
MaxHillebrand:
```suggestion
    "label": ["Alice, Daniel",
```

docs/using-wasabi/RPC.md
```diff
@@ -199,7 +255,8 @@ curl -s --data-binary '{"jsonrpc":"2.0","id":"1","method":"getwalletinfo"}' http
     "extendedAccountPublicKey": "tpubDCd1v6acjNY3uUqAtBGC6oBTGrCBWphMvkWjAqM2SFZahZb91JUTXZeZqxzscezR16XHkwi1723qo94EKgR75aoFaahnaHiiLP2JrrTh2Rk",
     "extendedAccountZpub": "vpub5YarnXR6ijVdw6G5mGhrUhf5bnodeCDJYtszFVW7LL3vr5HyRmJF8zfTZWzv6LjLPukmeR11ebWhLPLVVRjqbfyknJZdiwRWCyJcKeDdsC8",
     "accountKeyPath": "m/84'/0'/0'",
-    "masterKeyFingerprint": "323ec8d9"
+    "masterKeyFingerprint": "323ec8d9",
+    "balance": 0
```
MaxHillebrand: Why do you add these?

docs/using-wasabi/RPC.md
```diff
@@ -291,13 +349,13 @@ Now the mining fee will be subtracted from the output in which `subtractFee` was
 ( 0.4 - (mining fee) ) + 0.3 + 0.3
 
 ```bash
-curl -s --data-binary '{"jsonrpc":"2.0","id":"1","method":"send", "params": { "payments":[ {"sendto": "tb1qgvnht40a08gumw32kp05hs8mny954hp2snhxcz", "amount": 15000, "label": "David", "subtractFee": true }, {"sendto":"tb1qpyhfrpys6skr2mmnc35p3dp7zlv9ew4k0gn7qm", "amount": 86200, "label": "Michael"} ], "coins":[{"transactionid":"ab83d9d0b2a9873b8ab0dc48b618098f3e7fbd807e27a10f789e9bc330ca89f7", "index":0}], "feeTarget":2 }}' http://127.0.0.1:37128/ | jq
+curl -s --data-binary '{"jsonrpc":"2.0","id":"1","method":"send", "params": { "payments":[ {"sendto": "tb1qgvnht40a08gumw32kp05hs8mny954hp2snhxcz", "amount": 15000, "label": "David", "subtractFee": true }, {"sendto":"tb1qpyhfrpys6skr2mmnc35p3dp7zlv9ew4k0gn7qm", "amount": 86200, "label": "Michael"} ], "coins":[{"transactionid":"ab83d9d0b2a9873b8ab0dc48b618098f3e7fbd807e27a10f789e9bc330ca89f7", "index":0}], "feeTarget":2, "password": "password1234" }}' http://127.0.0.1:37128/ | jq
```
MaxHillebrand: So, is this correct for the current version of the RPC?

docs/using-wasabi/RPC.md
```diff
@@ -291,13 +349,13 @@ Now the mining fee will be subtracted from the output in which `subtractFee` was
...
 ```
 
 In case of error, it is reported in the json's error object:
 
 ```bash
-curl -s --data-binary '{"jsonrpc":"2.0","id":"1","method":"send", "params": { "payments": [{ "sendto": "tb1qnmfmkylkd548bbbcd9115b322891e27f741eb42c83ed982861ee121", "amount": 2015663, "label": "Mr. Who" }], "coins":[{"transactionid":"c68dacd548bbbcd9115b38ed982861ee121c5ef6e0f1022891e27f741eb42c83", "index":0}], "feeTarget": 2 }}' http://127.0.0.1:37128/ | jq
+curl -s --data-binary '{"jsonrpc":"2.0","id":"1","method":"send", "params": { "payments": [{ "sendto": "tb1qnmfmkylkd548bbbcd9115b322891e27f741eb42c83ed982861ee121", "amount": 2015663, "label": "Mr. Who" }], "coins":[{"transactionid":"c68dacd548bbbcd9115b38ed982861ee121c5ef6e0f1022891e27f741eb42c83", "index":0}], "feeTarget": 2, "password": "password1234" }}' http://127.0.0.1:37128/ | jq
```
MaxHillebrand: So, is this correct for the current version of the RPC?

docs/using-wasabi/RPC.md
PulpCattel: This is incorrect, the result for the label is a list `["Alice", "Daniel"]`

docs/using-wasabi/RPC.md
PulpCattel: The `getwalletinfo` method return also the balance.

docs/using-wasabi/RPC.md
PulpCattel: Yes, this is adding the missing parameter to the other examples. 
This is the `send` method and, as also lontivero explained, this accept the password (The `enqueue` is the method that doesn't accept it) and you can see that the first `send` example has the password parameter while the other two don't. So I added it to them too.

docs/using-wasabi/RPC.md
PulpCattel: same as above

PR #585, opened 2020-02-19
[BUILDING] give credits

MaxHillebrand:
This ready for review branch gives some more credits to projects quoted in the docs. Does anyone know of any further that should be credited?

```
- [Microsoft](https://dotnet.microsoft.com/download)
- [Qubes Operating System](https://qubes-os.org/doc/)
- [Whonix](https://whonix.org)
- [VirtualBox](https://virtualbox.org)
```

### DIFF
--- a/docs/building-wasabi/Credits.md
+++ b/docs/building-wasabi/Credits.md
@@ -8,8 +8,7 @@
 # Credits
 
 Part of the content in this documentation comes from other open source projects.
-Here are some credits, thanks and sources (in casual order):7
-
+Here are some credits, thanks and sources (in casual order):
 - [Bitcoin Wiki](https://en.bitcoin.it/wiki/Main_Page)
 - [BTC Pay Server Documentation](https://docs.btcpayserver.org/)
 - [Bisq Documentation](https://docs.bisq.network/)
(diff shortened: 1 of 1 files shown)


docs/building-wasabi/Credits.md
```diff
@@ -8,7 +8,8 @@
 # Credits
 
 Part of the content in this documentation comes from other open source projects.
-Here are some credits, thanks and sources (in casual order):
+Here are some credits, thanks and sources (in casual order):7
```
yahiheb:
```suggestion
Here are some credits, thanks and sources (in casual order):
```

docs/building-wasabi/Credits.md
```diff
@@ -8,7 +8,8 @@
...
+
```
yahiheb:
```suggestion
```

docs/building-wasabi/Credits.md
```diff
@@ -19,3 +20,7 @@ Here are some credits, thanks and sources (in casual order):
 - [CryptoSeb, Cryptopaper](https://github.com/cryptoseb/CryptoPaper)
 - [Tails](https://tails.boum.org/)
 - [BitSniff](https://github.com/m417z/bitsniff/)
+- [Microsoft](https://dotnet.microsoft.com/download)
```
yahiheb: I am not sure about this one.

MaxHillebrand: We actually do not describe how to install .NET, but only link to this download and let the Microsoft documentation speak for it self, thus I think it should be included.

yahiheb: yeah but I think we shouldn't include it (specially it is just a link to download .NET stuff), plus as you said we didn't copy any information from that source.

MaxHillebrand: Ah, now I understand your point, and yes, you are right!
```suggestion
```

PR #590, opened 2020-02-20
[USING] sending max with rpc

MaxHillebrand:
This ready for review branch adds a short explanation of how to send the maximum amount of a coin, by using `subtractFee`.


yahiheb:
> Now the mining fee will be subtracted from the output in which `subtractFee` was set to `true`

Just a question:
What if the fee is more than the amount of the output?

MaxHillebrand:
@yahiheb then this coin is dust, in the sense that it is uneconomical to be spent. Although theoretically it can be in the input of a valid transaction, there needs to be a second input, which pays for the missing fees of the size of that dust coin.

yahiheb:
> there needs to be a second input, which pays for the missing fees of the size of that dust coin.

So in this example `( 0.4 - (mining fee) ) + 0.3 + 0.3` if `0.4` is less than the `mining fee` then the other inputs will pay for the rest?

MaxHillebrand:
I think so @yahiheb , but I'm not sure...

yahiheb:
@lontivero

PR #595, opened 2020-02-21
Remove HWI from BIPs.md

RiccardoMasutti:


MaxHillebrand:
Well, it's not a BIP, but an industry standard...
I think it would be ok to leave it here. 
what do others say?

yahiheb:
Since it is the only one that is not a bip I say we remove it.
Or add a section where you can mention the supported industry standards that are not bips.

MaxHillebrand:
I think it would be better to have a separate section with industry standards.

PR #599, opened 2020-02-25
Fix JoinMarket capitalization.

nopara73:
Also testing CodeScene CI if it makes sense here.

### DIFF
--- a/docs/using-wasabi/Joinmarket.md
+++ b/docs/using-wasabi/Joinmarket.md
@@ -1,7 +1,7 @@
 ---
 {
-  "title": "Joinmarket",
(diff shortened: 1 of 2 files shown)


MaxHillebrand:
I'm actually not sure what the proper spelling is... the Joinmarket github says both `JoinMarket` and `Joinmarket` rather inconsistently.

nopara73:
image

MaxHillebrand:
I believe I got all the `Joinmarket` and made them upper case to `JoinMarket` [I used `git grep`].

Though I did not yet rename the file `using-wasabi/Joinmarket.md` to `using-wasabi/JoinMarket.md`, as this would break links. Should I change it too and fix all internal links?

nopara73:
> Should I change it too and fix all internal links?

No.

PR #613, opened 2020-03-10
[DECLUTTER FAQ] Add Backup and Recovery

RiccardoMasutti:
I created a new chapter, moved some FAQs and added new ones.

### DIFF
--- a/docs/FAQ/FAQ-UseWasabi.md
+++ b/docs/FAQ/FAQ-UseWasabi.md
@@ -70,30 +70,6 @@ A secure password manager software might also be used, but be careful here.
 ::::
 
 ::::details
-### How do I back up my mnemonic words?
-
-:::tip
-Write down your recovery words!
-:::
-
-Wasabi uses [BIP 39: mnemonic code for generating deterministic keys](/using-wasabi/BIPs.md#bip-39-mnemonic-code-for-generating-deterministic-keys) to enable easy backups of all private keys in the wallet.
-The mnemonic is displayed as 12 recovery words that are only shown once during the wallet generation.
-
-![](/WalletManagerRecoveryWords.png)
-
-:::danger
-In order to restore a wallet, you need BOTH the mnemonic words AND the password!
-:::
-
-It is a good idea to keep the two in separate analog backups, such as a laminated paper written with pencil.
-In order to defend against nature destroying fragile paper, you can consider stamping the words into metal.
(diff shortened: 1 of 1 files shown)


yahiheb:
I am not sure about adding all the reddit questions to the FAQs.
We have been doing this before but I think we should stop doing that.

RiccardoMasutti:
> I am not sure about adding all the reddit questions to the FAQs.
> We have been doing this before but I think we should stop doing that.

@MaxHillebrand What do you think?

yahiheb:
> @yahiheb I didn't even know that there were FAQs in reddit... I am in favor of having it all in the docs

I am not saying there are FAQs in Wasabi sub-reddit, what I am saying is we shouldn't add all the questions that get asked in reddit to the docs, otherwise we will have hundreds of questions which doesn't make sense.
FAQs stands for frequently asked questions not all the questions, and what lead us to this decluttering is that we add a lot of repeated information and some times not accurate or not clear info.

docs/FAQ/FAQ-UseWasabi.md
```diff
@@ -1032,6 +989,69 @@ Wasabi Wallet will use only the necessary coins to CoinJoin (up to 7 coins).
...
+Having a close guess of what the password is increases the possibility that you get your password.
+:::
+
+::::details
+### I lost seed words but still have .json file and the password. Is my wallet still recoverable?
+
+Yes, the .json file contains your private keys in it.
+That gives you access to your wallet in combination with the passphrase that you still have.
+
+In this case it is still advisable to create a new wallet from scratch and save the new seed + password.
```
yahiheb:
```suggestion
No.
The password you set is used as a 13th word (passphrase) as described in BIP39, you should back it up when you generate a wallet.
It is necessary to spend your bitcoin or to recover your wallet.
:::

:::details
### What should I do if I forget my password?

Try to use the [Password Finder Tool](/using-wasabi/PasswordFinder.md) to find your password.
Having a close guess of what the password is increases the possibility that you get your password.
:::

:::details
### I lost the recovery words but I still have the .json file and the password. Is my wallet still recoverable?

Yes, but in this case it is advisable to create a new wallet and back up the new recovery words and the password, then move your bitcoin there.

The .json file contains the encrypted secret that requires the password to derive your private keys.
That gives you access to your bitcoin.
```

PR #614, opened 2020-03-10
Add Tor Guard Node FAQ

RiccardoMasutti:


nopara73:
Where did I say this?

MaxHillebrand:
@nopara73 in person, while in the office, then I [and Lucas too] answered this question in the telegram group.

nopara73:
Please remove this. This is not a question anyone in his right mind would ever ask and I don't even think the answer (maybe even the question) is correct.

MaxHillebrand:
well... somebody actually did ask it... :laughing: 
But agreed, it's for sure not "frequently" asked.

yahiheb:
> Please remove this. This is not a question anyone in his right mind would ever ask and I don't even think the answer (maybe even the question) is correct.

> well... somebody actually did ask it... laughing
> But agreed, it's for sure not "frequently" asked.

This is why I have said [previously](https://github.com/zkSNACKs/WasabiDoc/pull/613#issuecomment-599212803) we shouldn't keep adding questions just for the sake of adding them, specially reddit or telegram questions.
We should be more selective on what should go in the docs, to avoid adding things and then removing them, as we have been doing lately.

PR #618, opened 2020-03-13
Update menu order and fix link

RiccardoMasutti:
Ordered in this way makes more sense.
Fixed a 404 link


yahiheb:
IMO the order is good as it is, address reuse should be number one on that list.

RiccardoMasutti:
@yahiheb I'd start with how Coins/input outputs works

MaxHillebrand:
I agree with @RiccardoMasutti, coins are so fundamental that they should be before address reuse [meaning to * **coins** *  have the same address]

yahiheb:
> [meaning to * coins * have the same address

How do you avoid having two coins in the same address if you don't know about address reuse?

From our documentation:
> The first rule of Bitcoin privacy is never reuse addresses!

> The first rule of Bitcoin privacy:
> 
> > Never reuse addresses!
> 
> The second rule of Bitcoin privacy:
> 
> > NEVER reuse addresses!!

docs/why-wasabi/README.md
```diff
@@ -21,9 +21,10 @@ And why Wasabi can make Bitcoin an even better tool of self-defense by default.
 - [Why Financial Privacy is Especially Important](/why-wasabi/WhyFinancialPrivacy.md)
 - [CoinJoin Legal Concern](/why-wasabi/LegalConcern.md)
 
-### Privacy in Bitcoin
-- [Address Reuse](/why-wasabi/AddressReuse.md)
+#### Privacy in Bitcoin
```
yahiheb, line 24:
```suggestion
### Privacy in Bitcoin
```

PR #619, opened 2020-03-13
Move FAQs from basics to advanced

RiccardoMasutti:
It seems to me more correct


yahiheb:
NACK. There is nothing advanced about those questions.

PR #623, opened 2020-03-13
Update TechnicalOverview.md

RiccardoMasutti:
Update link and remove obsolete sentence


yahiheb:
The medium article has good content, so I NACK that change.

Issue #582, opened 2020-02-19
How to check and fix broken links?


MaxHillebrand:
Is there some useful magic to be used to automatically check if the links are still working, and if possible, automatically fix them too?

RiccardoMasutti:
> Is there some useful magic to be used to automatically check if the links are still working, and if possible, automatically fix them too?

I take that. As soon as we implement Google Search Console, it will scan daily *.wasabiwallet.io and tell us crawl errors (broken links)

RiccardoMasutti:
Google Search Console still not implemented. I will make a new PR on /WalletWasabi repo to add a verification file

yahiheb:
> Google Search Console still not implemented.

Is there a PR for that?

RiccardoMasutti:
> > Google Search Console still not implemented.
> 
> Is there a PR for that?

Tomorrow I will

PeterXMR:
Some links still doesn't work ->
https://github.com/zkSNACKs/WasabiDoc/tree/master/docs/building-wasabi/
any update on that? I'll look on it if nobody is working on it.

yahiheb:
The work in the docs website.
https://docs.wasabiwallet.io/building-wasabi/

RiccardoMasutti:
@MaxHillebrand Now we receive alerts from GSC when Google crawls a 404 link/page.

yahiheb:
> @MaxHillebrand Now we receive alerts from GSC when Google crawls a 404 link/page.

Where do we receive that?

RiccardoMasutti:
> > @MaxHillebrand Now we receive alerts from GSC when Google crawls a 404 link/page.
> 
> Where do we receive that?

I personally reiceve them. I also added the rest of zksnacks' team as GSC administrators

PR #626, opened 2020-03-13
Fix 404 link WhyWasabi

RiccardoMasutti:


RiccardoMasutti:
https://github.com/zkSNACKs/WasabiDoc/issues/582

yahiheb:
Same as #618

yahiheb:
> Same as #618

@MaxHillebrand this shouldn't have been merged because it is the same in the referenced PR, and that is why you have conflicts in the other one.

MaxHillebrand:
Yes @yahiheb, but #618 is controversial, but this broken link fix is not. So I merged this, because if #618 does not end up getting merged, we still have the fix. 

Not sure if this is the proper way of handling it, as now there needs to be a conflict fix commit...

PR #653, opened 2020-03-20
Add wasabiHiddenService variable

RiccardoMasutti:
The goal is to have a single place to maintain the current Wasabi .onion as it is used in multiple places


yahiheb:
> How often does the hidden service change? Is the variable necessary here?

NACK.

PR #664, opened 2020-03-21
Remove misleading section

nopara73:


nopara73:
There was consensus on this change here: https://github.com/zkSNACKs/WasabiDoc/pull/661

PR #666, opened 2020-03-21
[DECLUTTER] Remove ToDo List

MaxHillebrand:
This ready for review PR removes the ToDo list file and any reference / link to it from the documentation.
As discussed in Slack and a bit in #661, we should not write promises about what might be included in the future - there is no roadmap.


MaxHillebrand:
Yes @nopara73 good catch - I forgot to stage the file before committing... It is removed now.

PR #688, opened 2020-03-30
Update README.md

Zolgarr:
Fixing a typo


Zolgarr:
Oh I see, I'll keep it in mind, thanks!

nopara73:
"The best block is no be there." - Miyagi

PR #707, opened 2020-04-04
[USING] add picture to mac install

MaxHillebrand:
This ready for review branch adds a picture of the MacOS warning that the image was downloaded from the [very very scary] internet.


yahiheb:
The picture is good, but its size is too big (2.37 MB).
I think it will be better to not put such big pictures, otherwise the repository will be very big.

MaxHillebrand:
Oh - You are right - I did not even see the size of it! 
Will follow up with a smaller one asap.

PR #710, opened 2020-04-04
fix picture size

MaxHillebrand:
A followup to #707 which decreases the size of the picture by 1.3 MB.


MaxHillebrand:
@yahiheb I did a screenshot same way that I did for all other of these pictures...

but maybe we can compress them all further.

PR #713, opened 2020-04-05
[USING] add Legal Documents

MaxHillebrand:
This ready for review branch imports the [terms and conditions, privacy notice, and legal statement](https://github.com/zkSNACKs/WalletWasabi/blob/master/WalletWasabi/Legal/Assets/LegalDocuments.txt). I did not change any of the wording, and please do not change anything in review neither, as the documentation should reflect exactly what is in the main wallet too. However, I did edit it to fit the layout, making it mark down compliant and pretty in vuepress, and also apply our writing conventions like one sentence per line.

Please review if I made any mistakes in the  markdown port.

### DIFF
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -227,8 +227,7 @@ module.exports = {
             "/using-wasabi/PasswordBestPractices.md",
             "/using-wasabi/BIPs.md",
             "/using-wasabi/IndustryStandards.md",
-            "/using-wasabi/TaC.md"
-
+            "/using-wasabi/LegalDocuments.md"
           ]
         }
       ],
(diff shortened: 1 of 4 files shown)


yahiheb:
Why do we want to import the legal documents to the docs website?

MaxHillebrand:
@yahiheb so that users can read it before downloading the package.
It is also mentioned in the ToC that the text will be available on the website

yahiheb:
Why not simply send them to https://github.com/zkSNACKs/WalletWasabi/blob/master/WalletWasabi/Legal/Assets/LegalDocuments.txt
I don't think this is the way to do it. Plus we have to maintain this.

MaxHillebrand:
Yes, good call @yahiheb, thanks.

PR #723, opened 2020-04-09
Fix 404 links

RiccardoMasutti:


docs/why-wasabi/TransactionSurveillanceCompanies.md
```diff
@@ -112,3 +111,4 @@ Whether the activities of transaction surveillance company actually help against
 - [BlockSeer](https://www.blockseer.com/)
 - [ScoreChain](https://www.scorechain.com/)
 - [Crystal](https://crystalblockchain.com/)
+- Coinvalidation
```
MaxHillebrand, line 114: Got a link here too?
https://www.coinvalidation.com/ seems to be down...

RiccardoMasutti: That's why I removed it

MaxHillebrand: Is the company still active? Should we remove them altogether then?

RiccardoMasutti: > Is the company still active? Should we remove them altogether then?

Can't find info about it...

MaxHillebrand: They started back in 2013, but I don't think they are still around... 

well, good to keep the historical record. the PR is good.

PR #733, opened 2020-04-10
[USING] rework Send

MaxHillebrand:
This ready for review branch makes a minor rework of the Send chapter.

### DIFF
--- a/docs/using-wasabi/Send.md
+++ b/docs/using-wasabi/Send.md
@@ -16,7 +16,7 @@
 1. Select the coins you want to spend.
 2. Specify a destination address.
 3. Label the observers of this transaction.
-4. Specify the amount to send to the destination address, press the `Max` button to send a whole coin without change.
+4. Specify the amount to send to the destination address, or press the `Max` button to send the whole selected amount.
 5. Select a mining fee.
 6. Enter the password.
 7. Click `Send Transaction`.
--- a/docs/using-wasabi/Send.md
+++ b/docs/using-wasabi/Send.md
@@ -15,8 +15,8 @@
 
 1. Select the coins you want to spend.
 2. Specify a destination address.
-3. If necessary label the observers of this transaction.
-4. Specify the amount to send to the destination address, press the `Max` button to send the whole coins.
+3. Label the observers of this transaction.
+4. Specify the amount to send to the destination address, press the `Max` button to send a whole coin without change.
 5. Select a mining fee.
 6. Enter the password.
 7. Click `Send Transaction`.
@@ -26,8 +26,8 @@
 ## Coins
 
 A coin is an unspent transaction output (UTXO), a chunk of bitcoin which can be sent in a future transaction.
-In the Wasabi wallet `Send` tab, you see a list of all the coins you can spend because you have their private keys.
-You can get a coin by first [receiving](/using-wasabi/Receive.md) them from someone else, for example by earning them or buying them for fiat currency.
+In the Wasabi wallet `Send` tab, you see a list of all the coins you can spend because.
+You can get a coin by first [receiving](/using-wasabi/Receive.md) them from someone else, for example by earning them or exchanging them for fiat currency.
 You can spend one or more coins by selecting them in the `Send` tab, if your payment amount is below the value of the selected coins, then you will receive a [change coin](/using-wasabi/ChangeCoins.md) back.
 
 ## Clusters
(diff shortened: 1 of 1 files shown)


docs/using-wasabi/Send.md
```diff
@@ -74,17 +75,21 @@ If it is below the value of the selected inputs, then the leftover value will be
 You can send a whole coin by selecting the `Max` button, which will build a transaction with only one output, the receiving address, and no change.
 You can also see the current US Dollar value of the sending amount.
 
-If you specify a rounded amount, like `0.01000000 bitcoin`, then the change output will not be rounded, like `0.08968413 bitcoin`.
+:::tip Avoid sending rounded values
+If you specify a rounded amount, like `0.0100 0000 bitcoin`, then the change output will not be rounded, like `0.0896 8413 bitcoin`.
```
RiccardoMasutti, line 79:
```suggestion
If you specify a rounded amount, like `0.01000000 bitcoin`, then the change output will not be rounded, like `0.0896 8413 bitcoin`.
```

docs/using-wasabi/Send.md
```diff
@@ -74,17 +75,21 @@ If it is below the value of the selected inputs, then the leftover value will be
...
 This makes it easy for an observer to conclude that the spending amount was the `0.01 bitcoin`, and that the other output is the change back to the sender.
-So in order to increase your privacy, you can set a non-rounded amount, like `0.01016843`.
+So in order to increase your privacy, you can set a non-rounded amount, like `0.0101 6843`.
```
RiccardoMasutti, line 81:
```suggestion
So in order to increase your privacy, you can set a non-rounded amount, like `0.01016843`.
```

docs/using-wasabi/Send.md
```diff
@@ -15,8 +15,8 @@
 
 1. Select the coins you want to spend.
 2. Specify a destination address.
-3. If necessary label the observers of this transaction.
-4. Specify the amount to send to the destination address, press the `Max` button to send the whole coins.
+3. Label the observers of this transaction.
+4. Specify the amount to send to the destination address, press the `Max` button to send a whole coin without change.
```
yahiheb, line 19: In step 1 we say `Select the coins`.
```suggestion
4. Specify the amount to send to the destination address, or press the `Max` button to send the whole coins.
```

docs/using-wasabi/Send.md
```diff
@@ -26,8 +26,8 @@
 ## Coins
 
 A coin is an unspent transaction output (UTXO), a chunk of bitcoin which can be sent in a future transaction.
-In the Wasabi wallet `Send` tab, you see a list of all the coins you can spend because you have their private keys.
-You can get a coin by first [receiving](/using-wasabi/Receive.md) them from someone else, for example by earning them or buying them for fiat currency.
+In the Wasabi wallet `Send` tab, you see a list of all the coins you can spend because.
```
yahiheb, line 29:
```suggestion
In the Wasabi wallet `Send` tab, you see a list of all the coins you can spend.
```

docs/using-wasabi/Send.md
```diff
@@ -40,13 +40,14 @@ The goal is to know the observers who know about your coins and try to reduce th
 
 ## Anonymity Set
 
+When you receive a bitcoin in a regular transaction with one input, and two outputs, then your coin can be linked to this one output.
```
yahiheb, line 43: You mean when you send?

docs/using-wasabi/Send.md
```diff
@@ -55,17 +56,17 @@ By default they have an anonymity set of `2`, `21` and `50`, however, this can b
 
 When sending bitcoin, you need to know the destination address of the receiver.
 This commits to the spending condition that the receiver agrees to have for this coin.
-The address can be a public key hash [starting with 1...], a script hash [starting with 3...], or a native segwit bech32 public key hash [starting with bc1q...].
+The address can be a public key hash [starting with `1...`], a script hash [starting with `3...`], or a native segwit bech32 public key hash [starting with `bc1q...`].
```
yahiheb, line 59:
```suggestion
The address can be a public key hash [starting with `1`], a script hash [starting with `3`], or a native segwit bech32 public key hash [starting with `bc1q`].
```

docs/using-wasabi/Send.md
MaxHillebrand: I think this is better readable... [it should be [in the wallet too](https://github.com/zkSNACKs/WalletWasabi/issues/2588), but that's OT]

docs/using-wasabi/Send.md
```diff
@@ -115,3 +120,6 @@ Afterwards the password is wiped from the memory.
 
 Once the transaction is signed, Wasabi will connect to a random Bitcoin P2P node over Tor and provide this transaction, then it will immediately disconnect.
 This first node will gossip the transaction throughout the network, then miners can include it in a block.
+
+If for some reason the first broadcast fails, then if you have Wasabi connected to [your own Bitcoin full node](/using-wasabi/BitcoinFullNode.md) this node will broadcast the transaction to the network.
+If this also fails, or you don't have a full node, then the transaction is send to the backend coordinator with a new Tor identity, who then gossips the transaction to the network.
```
yahiheb, line 125:
```suggestion
If this also fails, then the transaction is sent to the backend coordinator with a new Tor identity, who then broadcasts the transaction to the network.
```

docs/using-wasabi/Send.md
MaxHillebrand: Yes?
but it doesn't matter which coins you have selected, pressing `Max` will send "a whole coin" [which might be confusing] without a change coin [to be more clear]

docs/using-wasabi/Send.md
MaxHillebrand: Well both...

But you are right, the example is more precise when sending / receiving a change.

```suggestion
When you send a bitcoin in a regular transaction with one input, and two outputs, then your change coin can be linked to this one input.
```

docs/using-wasabi/Send.md
yahiheb: I think we should be talking about the selected coins not the output coin here.
Imagine if you were sending to many that sentence wouldn't be correct.

MaxHillebrand: Even in P2M, when the Max button is pressed, there is no change back to the user wallet, but the reminder goes to one of the destination outputs.

yahiheb: How about:
```suggestion
4. Specify the amount to send to the destination address, or press the `Max` button to send the whole selected amount.
```

PR #735, opened 2020-04-10
[USING] rework Bitcoin Full Node

MaxHillebrand:
This ready for review branch reworks the chapter of Bitcoin Full Node.

The main addition is a paragraph that the future goal is to create the block filters with the node.

### DIFF
--- a/docs/using-wasabi/BitcoinFullNode.md
+++ b/docs/using-wasabi/BitcoinFullNode.md
@@ -45,7 +45,7 @@ Wasabi also queries if your local mempool has unconfirmed transactions that are 
 Further, your full node is used to estimate the current mining fee level based on the bitcoind `smartfee` algorithm.
 Your full node is not used to broadcast your transactions, as this is done through a random peer-to-peer node with a new tor identity, which is better for your [network level privacy](/why-wasabi/NetworkLevelPrivacy.md).
 
-The future goal is that the local full node is used to generate the BIP 158 block filters from the verified blocks.
+The future goal is that the local full node (Knots) is used to generate the BIP 158 block filters from the verified blocks.
 Only then is there full verification, and no trust in the zkSNACKs server.
 
 ## bitcoind within Wasabi
--- a/docs/using-wasabi/BitcoinFullNode.md
+++ b/docs/using-wasabi/BitcoinFullNode.md
@@ -31,7 +31,7 @@ There are several software implementations that function as a full node, for exa
 
 ## How does Wasabi use your Bitcoin full node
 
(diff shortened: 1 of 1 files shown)


docs/using-wasabi/BitcoinFullNode.md
```diff
@@ -31,8 +31,8 @@ There are several software implementations that function as a full node, for exa
 
 ## How does Wasabi use your Bitcoin full node
 
-Wasabi currently has a work in progress partial Bitcoin Knots integration.
-The zkSNACKs coordinating server broadcasts [BIP 158 block filters](/using-wasabi/BIPs.md#bip-158-compact-block-filters-for-light-clients) to all Wasabi clients, who locally check if the filter hits for their public keys.
+Wasabi currently has a work in progress partial Bitcoin full node integration.
```
yahiheb, line 34: IMO we should specify `Knots` here, most people might think it is `Core`.

MaxHillebrand: Well, my thinking is that the **integration** is a full node [for example, can be Core OR Knots on remote computer] but the **built-in** is Knots specifically.

yahiheb: The integration is not any full node, it is knots specifically.
The work in progress will allow the built-in Knots to generate the block filters, which is not true for Core.

MaxHillebrand: Yes, this is true, but for a long time now, Core is used for filters / fee estimation / mempool etc. 

the verification is only for Knots though.

But I get your point, and I'll commit and merge it.

MaxHillebrand:
```suggestion
Wasabi currently has a work in progress partial Bitcoin Knots integration.
```

docs/using-wasabi/BitcoinFullNode.md
```diff
@@ -45,6 +45,9 @@ Wasabi also queries if your local mempool has unconfirmed transactions that are
 Further, your full node is used to estimate the current mining fee level based on the bitcoind `smartfee` algorithm.
 Your full node is not used to broadcast your transactions, as this is done through a random peer-to-peer node with a new tor identity, which is better for your [network level privacy](/why-wasabi/NetworkLevelPrivacy.md).
 
+The future goal is that the local full node is used to generate the BIP 158 block filters from the verified blocks.
```
yahiheb, line 48: I missed this before, I think we should precise that we are talking about Knots here also.
```suggestion
The future goal is that the local full node (Knots) is used to generate the BIP 158 block filters from the verified blocks.
```

PR #738, opened 2020-04-11
[DECLUTTER USING] remove Deanonimization

MaxHillebrand:
This ready for review branch removes the chapter on Deanonymization. It is all over the place, and I don't know how to fix it. Most of the information in here is mentioned elsewhere in the docs, it doesn't add anything new, and it's not really about Wasabi at all.


MaxHillebrand:
Why @RiccardoMasutti?
Please provide a reasoning.

RiccardoMasutti:
Can you provide me where these deanonymization tips are mentioned elsewhere in the docs?

MaxHillebrand:
The whole documentation is a treatise on "how to not get deanonymized", so having this chapter to discuss an arbitrarily chosen subsection of it seems odd.

> Revealing your Bitcoin address before it goes into the blockchain could let others track you

Leaking public keys is mentioned many times.

> Possessing a wallet file might be enough proof that you control Bitcoin

That the public keys are unencrypted in the wallet file is mentioned too.

> Be careful with public spaces

Address reuse / publicly advertising donation addresses is mentioned too.

yahiheb:
I am not sure this should be removed.

> Most of the information in here is mentioned elsewhere in the docs, it doesn't add anything new

What it adds is a stand alone chapter for deanonimization, so you don't have to read all the docs to get that information.

> and it's not really about Wasabi at all.

There are many other information that are not about Wasabi in the docs.

MaxHillebrand:
I think it should either by entirely reworked, or removed. I started to re-write it, but then found the topic near impossible to approach. I tried to change the title to `Linking Pseudonymous Identities` but then we're again in coin control and utxo model. 

>  so you don't have to read all the docs to get that information.

This is a good point @yahiheb, but I have no clue what information should be in such a chapter...

yahiheb:
Why edit the title from something easy to understand `Deanonimization` to something more complicated `Linking Pseudonymous Identities`?

We should use less complicated terms to explain things.

PR #742, opened 2020-04-11
[USING] rework password finder

MaxHillebrand:
This ready for review branch reworks the password finder. Mainly a clarification on the brute force protection of BIP38, with minor layout fixes.

### DIFF
--- a/docs/using-wasabi/PasswordFinder.md
+++ b/docs/using-wasabi/PasswordFinder.md
@@ -16,18 +16,18 @@ This tool tries to find the password that decrypts the encrypted secret key stor
 
 ## Limitations
 
-Wasabi Wallet protects the encrypted secret key with with a user specified password ([BIP 38](/using-wasabi/BIPs.md#bip-38-password-protected-private-key)).
+Wasabi Wallet protects the encrypted secret key with a user specified password ([BIP 38](/using-wasabi/BIPs.md#bip-38-password-protected-private-key)).
 One of the features of this method is that it has brute force protection, meaning even if the password is wrong, it still generates a valid (but unused) private key.
 However, the Wasabi Password Finder has the wallet file, and thus it knows the used public keys of the wallet, and it can check the generated private key against the known to be valid public key.
 It is important to know that Wasabi Password Finder is not for breaking wallet passwords but for finding minor errors (typos) in a roughly known password. 
 
 ## Usage
 
-#### Windows
+### Windows
 
 To use Wasabi's command line tools on Windows you have to use `wassabeed.exe` that is inside your `Program Files\WasabiWallet`.
 
-#### Linus and macOS
+### Linus and macOS
 
 Use the `wassabee` command.
 
@@ -47,7 +47,7 @@ eg: ./wassabee findpassword --wallet:MyWalletName --numbers:false --symbold:true
   -h, --help                 Show Help
 ```
 
-For example, there is a called `MagicalCryptoWallet.json`, the correct password is `pass`, but the user thinks it is `pasd`.
+For example, there is a wallet called `MagicalCryptoWallet.json`, the correct password is `pass`, but the user thinks it is `pasd`.
 
 ```
 $ wassabee findpassword --wallet:MagicalCryptoWallet
--- a/docs/using-wasabi/PasswordFinder.md
+++ b/docs/using-wasabi/PasswordFinder.md
@@ -16,18 +16,23 @@ This tool tries to find the password that decrypts the encrypted secret key stor
 
 ## Limitations
 
-Wasabi Wallet protects the encrypted secret key with the same technology used to protect paper wallets ([BIP 38](/using-wasabi/BIPs.md#bip-38-password-protected-private-key)) and for that reason it is computationally infeasible to brute force the password using all the possible combinations.
-It is important to know that Wasabi Password Finder is not for breaking wallet passwords but for finding errors (typos) in an already known password. 
+Wasabi Wallet protects the encrypted secret key with with a user specified password ([BIP 38](/using-wasabi/BIPs.md#bip-38-password-protected-private-key)).
+One of the features of this method is that it has brute force protection, meaning even if the password is wrong, it still generates a valid [but unused] private key.
+However, the Wasabi Password Finder has the wallet file, and thus it knows the used public keys of the wallet, and it can check the generated private key against the known to be valid public key.
+It is important to know that Wasabi Password Finder is not for breaking wallet passwords but for finding minor errors (typos) in a roughly known password. 
 
 ## Usage
 
-To use Wasabi's command line tools on Windows you have to use `wassabeed.exe` that is inside your `Program Files\WasabiWallet`.
-On Linux and macOS you can use the same software that you use for launching the GUI (`wassabee`).
+#### Windows
 
-Let us start giving a glance to the command `help`:
+To use Wasabi's command line tools on Windows you have to use `wassabeed.exe` that is inside your `Program Files\WasabiWallet`.
+
+#### Linus and macOS
+
+Use the `wassabee` command.
 
 ```
-$ wassabee run help findpassword
+$ wassabee help findpassword
 usage: findpassword --wallet:WalletName --language:lang --numbers:[TRUE|FALSE] --symbold:[TRUE|FALSE]
 
 Tries to find typing mistakes in the user password by brute forcing it char by char.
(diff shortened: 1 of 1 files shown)


docs/using-wasabi/PasswordFinder.md
```diff
@@ -16,18 +16,23 @@ This tool tries to find the password that decrypts the encrypted secret key stor
 
 ## Limitations
 
-Wasabi Wallet protects the encrypted secret key with the same technology used to protect paper wallets ([BIP 38](/using-wasabi/BIPs.md#bip-38-password-protected-private-key)) and for that reason it is computationally infeasible to brute force the password using all the possible combinations.
-It is important to know that Wasabi Password Finder is not for breaking wallet passwords but for finding errors (typos) in an already known password. 
+Wasabi Wallet protects the encrypted secret key with with a user specified password ([BIP 38](/using-wasabi/BIPs.md#bip-38-password-protected-private-key)).
+One of the features of this method is that it has brute force protection, meaning even if the password is wrong, it still generates a valid [but unused] private key.
```
RiccardoMasutti, line 20:
```suggestion
One of the features of this method is that it has brute force protection, meaning even if the password is wrong, it still generates a valid (but unused) private key.
```

docs/using-wasabi/PasswordFinder.md
```diff
@@ -16,18 +16,23 @@ This tool tries to find the password that decrypts the encrypted secret key stor
 
 ## Limitations
 
-Wasabi Wallet protects the encrypted secret key with the same technology used to protect paper wallets ([BIP 38](/using-wasabi/BIPs.md#bip-38-password-protected-private-key)) and for that reason it is computationally infeasible to brute force the password using all the possible combinations.
-It is important to know that Wasabi Password Finder is not for breaking wallet passwords but for finding errors (typos) in an already known password. 
+Wasabi Wallet protects the encrypted secret key with with a user specified password ([BIP 38](/using-wasabi/BIPs.md#bip-38-password-protected-private-key)).
```
yahiheb, line 19:
```suggestion
Wasabi Wallet protects the encrypted secret key with a user specified password ([BIP 38](/using-wasabi/BIPs.md#bip-38-password-protected-private-key)).
```

docs/using-wasabi/PasswordFinder.md
```diff
@@ -16,18 +16,23 @@ This tool tries to find the password that decrypts the encrypted secret key stor
...
+One of the features of this method is that it has brute force protection, meaning even if the password is wrong, it still generates a valid (but unused) private key.
+However, the Wasabi Password Finder has the wallet file, and thus it knows the used public keys of the wallet, and it can check the generated private key against the known to be valid public key.
+It is important to know that Wasabi Password Finder is not for breaking wallet passwords but for finding minor errors (typos) in a roughly known password. 
 
 ## Usage
 
+#### Windows
```
yahiheb, line 26:
```suggestion
### Windows
```

docs/using-wasabi/PasswordFinder.md
```diff
@@ -16,18 +16,23 @@ This tool tries to find the password that decrypts the encrypted secret key stor
...
+
 To use Wasabi's command line tools on Windows you have to use `wassabeed.exe` that is inside your `Program Files\WasabiWallet`.
-On Linux and macOS you can use the same software that you use for launching the GUI (`wassabee`).
 
-Let us start giving a glance to the command `help`:
+#### Linus and macOS
```
yahiheb, line 30:
```suggestion
### Linus and macOS
```

docs/using-wasabi/PasswordFinder.md
```diff
@@ -42,15 +47,14 @@ eg: ./wassabee findpassword --wallet:MyWalletName --numbers:false --symbold:true
   -h, --help                 Show Help
 ```
 
-Now, let us find a typo in a wallet called `MagicalCryptoWallet.json`.
-For the sake of the example let us say I have created this wallet and I think the password is `pasd` but it was created with the password `pass` by accident.
+For example, there is a called `MagicalCryptoWallet.json`, the correct password is `pass`, but the user thinks it is `pasd`.
```
yahiheb, line 50:
```suggestion
For example, there is a wallet called `MagicalCryptoWallet.json`, the correct password is `pass`, but the user thinks it is `pasd`.
```

docs/using-wasabi/PasswordFinder.md
```diff
@@ -16,18 +16,23 @@ This tool tries to find the password that decrypts the encrypted secret key stor
...
-On Linux and macOS you can use the same software that you use for launching the GUI (`wassabee`).
 
-Let us start giving a glance to the command `help`:
+### Linus and macOS
+
+Use the `wassabee` command.
 
 ```
-$ wassabee run help findpassword
+$ wassabee help findpassword
```
yahiheb, line 35: Is 'run' not necessary here?

yahiheb:
@lontivero can you take a look at this.

docs/using-wasabi/PasswordFinder.md
MaxHillebrand: run breaks it.
Tested on Qubes 4 Debian 10 - not sure about Windows though

docs/using-wasabi/PasswordFinder.md
```diff
@@ -16,18 +16,23 @@ This tool tries to find the password that decrypts the encrypted secret key stor
 
 ## Limitations
 
-Wasabi Wallet protects the encrypted secret key with the same technology used to protect paper wallets ([BIP 38](/using-wasabi/BIPs.md#bip-38-password-protected-private-key)) and for that reason it is computationally infeasible to brute force the password using all the possible combinations.
-It is important to know that Wasabi Password Finder is not for breaking wallet passwords but for finding errors (typos) in an already known password. 
+Wasabi Wallet protects the encrypted secret key with a user specified password ([BIP 38](/using-wasabi/BIPs.md#bip-38-password-protected-private-key)).
+One of the features of this method is that it has brute force protection, meaning even if the password is wrong, it still generates a valid (but unused) private key.
```
lontivero, line 20: There is a deep misunderstanding of bip38 here. Bip38 is simply a method to encrypt a private key and encode it as a printable string. This method involves the usage of scrypt to make the brute force attacks harder. 

Any attempt to decrypt the encrypted private key with the wrong password results in an error. It does never return an incorrect private key.

Bip39 is the method to generate a seed from human-readable words. It allows to add additional entropy by using extra words, what we usually refer as the 13th or 25th word. This means that any change any of the words will generate a completely different seed. For example, imagine we don't provide a 13th word but we make a mistake writing the 12 words in our piece of paper, it can be a typo, a mistake in the order of the words or whatever, that will result in a different seed.

In wasabi both bips are in place and they both share something in common: the 13th word used in bip39 is the password used in the encryption (bip38). But brute force protection (bip38) and seed derivation (bip39) are really two very different things.

docs/using-wasabi/PasswordFinder.md
```diff
@@ -16,18 +16,23 @@ This tool tries to find the password that decrypts the encrypted secret key stor
...
+However, the Wasabi Password Finder has the wallet file, and thus it knows the used public keys of the wallet, and it can check the generated private key against the known to be valid public key.
```
MaxHillebrand, line 21:
```suggestion
The Wasabi Password finder requires the wallet file.
```

@lontivero how about this?

yahiheb: How about we simply keep that part as it was.

MaxHillebrand: Because I think this is not accurate... If it is computationally infeasible depends on the password chosen, not on the BIP38 standard per se...
` and for that reason it is computationally infeasible to brute force the password using all the possible combinations.`

yahiheb: we can simply add `(assuming a secure password was chosen)`.
No need to completely re-write all that.

docs/using-wasabi/PasswordFinder.md
```diff
@@ -16,18 +16,23 @@ This tool tries to find the password that decrypts the encrypted secret key stor
...
+It is important to know that Wasabi Password Finder is not for breaking wallet passwords but for finding minor errors (typos) in a roughly known password.
```
yahiheb, line 22:
```suggestion
Wasabi Wallet protects the encrypted secret key with the same technology used to protect paper wallets ([BIP 38](/using-wasabi/BIPs.md#bip-38-password-protected-private-key)) and for that reason it is computationally infeasible to brute force the password using all the possible combinations. (assuming a secure password was chosen)
It is important to know that Wasabi Password Finder is not for breaking wallet passwords but for finding errors (typos) in an already known password.
```

PR #748, opened 2020-04-12
[USING] add cd to daemon

MaxHillebrand:
This ready for review branch fixes some uncertainty of the daemon chapter, based on user reports.

mainly, it adds the `cd` instructions for macOS and Linux. @yahiheb, how do you change directory in Windows?

There are still reports it doesn't work with `wassabee mix` on Mac, even when in the Application directory. So do not yet merge, still trouble shooting.

### DIFF
--- a/docs/using-wasabi/Daemon.md
+++ b/docs/using-wasabi/Daemon.md
@@ -43,6 +43,7 @@ If the package is installed, execute in the command line in any directory:
 If the source code is built, navigate to the WalletWasabi.Gui folder (inside the cloned repo) and execute:
 
 ```bash
+~$ cd ~/WalletWasabi/WalletWasabi.Gui
 ~/WalletWasabi/WalletWasabi.Gui$ dotnet run -- mix --wallet:MyFirstWallet --destination:MySecondWallet --keepalive
 ```
 
@@ -51,12 +52,14 @@ If the source code is built, navigate to the WalletWasabi.Gui folder (inside the
 If the package is installed, navigate to the Applications directory and execute: 
 
 ```bash
+~$ cd ~/Applications/Wasabi\ Wallet.app/Contents/MacOs
 ~/Applications/Wasabi\ Wallet.app/Contents/MacOs $ wassabee mix --wallet:MyFirstWallet --destination:MySecondWallet --keepalive
 ```
 
 If the source code is built, navigate to the WalletWasabi.Gui folder (inside the cloned repo) and execute:
 
 ```bash
+~$ cd ~/WalletWasabi/WalletWasabi.Gui
 ~/WalletWasabi/WalletWasabi.Gui$ dotnet run -- mix --wallet:MyFirstWallet --destination:MySecondWallet --keepalive
 ```
 
--- a/docs/using-wasabi/Daemon.md
+++ b/docs/using-wasabi/Daemon.md
@@ -40,7 +40,7 @@ If the package is installed, execute in the command line in any directory:
 ~$ wassabee mix --wallet:MyFirstWallet --destination:MySecondWallet --keepalive
 ```
 
-If the source code is built, navigate to the WalletWasabi.Gui folder (inside the cloned repo) and execute:
+If the source code is built:
 
 ```bash
 ~$ cd ~/WalletWasabi/WalletWasabi.Gui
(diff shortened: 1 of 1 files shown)


MaxHillebrand:
User reports that what was previously described does not work, only `wassabee`:
`~/Applications/Wasabi\ Wallet.app/Contents/MacOs$ wassabee mix --wallet:MyFirstWallet --destination:MySecondWallet --keepalive`

But, that this PR does work, when it is `./wassabee`:
`~/Applications/Wasabi\ Wallet.app/Contents/MacOs$ ./wassabee mix --wallet:MyFirstWallet --destination:MySecondWallet --keepalive`

yahiheb:
I NACK adding the `cd` command, because the line above it says `navigate to the WalletWasabi.Gui folder and execute`, it doesn't make sense to execute the `cd` command after you navigate to the folder.

docs/using-wasabi/Daemon.md
```diff
@@ -43,6 +43,7 @@ If the package is installed, execute in the command line in any directory:
 If the source code is built, navigate to the WalletWasabi.Gui folder (inside the cloned repo) and execute:
```
MaxHillebrand, line 43:
```suggestion
If the source code is built:
```

@yahiheb how about it now?

docs/using-wasabi/Daemon.md
```diff
@@ -51,12 +52,14 @@ If the source code is built, navigate to the WalletWasabi.Gui folder (inside the
 If the package is installed, navigate to the Applications directory and execute:
```
MaxHillebrand, line 52:
```suggestion
If the package is installed:
```

docs/using-wasabi/Daemon.md
```diff
@@ -51,12 +52,14 @@ If the source code is built, navigate to the WalletWasabi.Gui folder (inside the
 If the package is installed, navigate to the Applications directory and execute: 
 
 ```bash
-~/Applications/Wasabi\ Wallet.app/Contents/MacOs $ wassabee mix --wallet:MyFirstWallet --destination:MySecondWallet --keepalive
+~$ cd ~/Applications/Wasabi\ Wallet.app/Contents/MacOs
+~/Applications/Wasabi\ Wallet.app/Contents/MacOs$ ./wassabee mix --wallet:MyFirstWallet --destination:MySecondWallet --keepalive
 ```
 
 If the source code is built, navigate to the WalletWasabi.Gui folder (inside the cloned repo) and execute:
```
MaxHillebrand, line 59:
```suggestion
If the source code is built:
```

molnard:
macOS:

1. I did not use the ~ character.
2. You do not need to navigate anywhere, just enter the command mentioned here: https://github.com/zkSNACKs/WalletWasabi/issues/3525#issuecomment-612775860

windows:

1. Go to the terminal(cmd).
2. Enter: `"c:\Program Files\WasabiWallet\wassabeed.exe" -v`

MaxHillebrand:
Can we merge @yahiheb ?

docs/using-wasabi/Daemon.md
```diff
@@ -40,23 +40,26 @@ If the package is installed, execute in the command line in any directory:
...
 
 ### MacOS
 
-If the package is installed, navigate to the Applications directory and execute: 
+If the package is installed:
 
 ```bash
-~/Applications/Wasabi\ Wallet.app/Contents/MacOs $ wassabee mix --wallet:MyFirstWallet --destination:MySecondWallet --keepalive
+~$ cd ~/Applications/Wasabi\ Wallet.app/Contents/MacOs
+~/Applications/Wasabi\ Wallet.app/Contents/MacOs$ ./wassabee mix --wallet:MyFirstWallet --destination:MySecondWallet --keepalive
```
molnard, line 56:
```suggestion
/Applications/Wasabi\ Wallet.app/Contents/MacOS/wassabee mix --wallet:MyFirstWallet --destination:MySecondWallet --keepalive
```

MaxHillebrand: This would work too, but then when restarting the daemon, the user needs to execute the long path command all the time again. If you change directory once, then you only need `./wassabee`...

molnard: or just press the up button and the same command will appear.

molnard: I am OK with both if it solves the user problems.

MaxHillebrand: yes - pressing up works too. 
I'm not sure what is better, both work of course. 
@yahiheb do you have any idea?

molnard: I would prefer the one line solution on macOS

yahiheb: IMO we should not merge this and simply keep it as it was.
Users who want to use the daemon should be able to navigate to the folder and execute the command to mix.

MaxHillebrand: I have heard questions from two different users who got confused with the current version...

so, either, it should be David's suggestion, or my PR, I don't think the current one is good.

PR #749, opened 2020-04-13
Remove duplicate license

yahiheb:


RiccardoMasutti:
I agree with Max

yahiheb:
> I think the license should be in the rendered documentation too

What is the reason for that? I really don't see why it should be there.
Also you need to maintain it separately .

MaxHillebrand:
It should be there so that users can read it inside the documentation, without going either to the "confusing" github UI, or to the "official" license website. 

I don't think maintenance is an issue though, the license doesn't change, only the year, and that'll be easily done once a year...

lontivero:
I agree with @yahiheb, it is just one link.

PR #751, opened 2020-04-13
fixing minor grammar and typos

Zolgarr:


docs/why-wasabi/LegalConcern.md
```diff
@@ -11,8 +11,8 @@
 
 ---
 
-Using Bitcoin and other cryptocurrencies in the everyday course of businesses is becoming more and more popular.
-Digital money has a lot of advantages, from unstoppable transactions, to low fees, and instant execution of international payments.
+Using Bitcoin and other cryptocurrencies in the everyday course of businesses are becoming more and more popular.
```
yahiheb, line 14: I think `is` is correct here but I am not sure.
How about something like this:
```suggestion
Bitcoin and other cryptocurrencies usage in the everyday course of businesses is becoming more and more popular.
```

MaxHillebrand: It is `is`, not `are`.

docs/why-wasabi/LegalConcern.md
```diff
@@ -71,12 +71,12 @@ Other competitors can now demand similar discounts, forcing industries into a vi
 As each and every price discount is negotiated on a case by case basis, considering factors as creditworthiness and timing of payment, transparent supplier prices would only cause enormous pressure on suppliers to deal with negotiations.
 
 Beyond the price of products, the timing of payments is also fully transparent on the blockchain.
-However, in certain circumstances, it is also a crucial business secret to keep the information of the sequence of invoice payments indoors.
+However, in certain circumstances, it is also a crucial business secret to keep the information on the sequence of invoice payments indoors.
```
MaxHillebrand, line 74: I think `of` is correct here?

docs/why-wasabi/LegalConcern.md
Zolgarr: Indeed, you're right, my bad!

docs/why-wasabi/LegalConcern.md
Zolgarr: I think `of` works too, but I think it's more appropriate to use `on` or `about`
When using `about`, then `the` should be left out:
"To keep information about the sequence of invoice payments indoors."

PR #753, opened 2020-04-13
Remove testnet question

yahiheb:
Most of what is said there is already in this chapter
https://docs.wasabiwallet.io/using-wasabi/Testnet.html

### DIFF
--- a/docs/FAQ/FAQ-UseWasabi.md
+++ b/docs/FAQ/FAQ-UseWasabi.md
@@ -1256,25 +1256,6 @@ You can use the [Wasabi RPC server `gethistory` call](/using-wasabi/RPC.md#gethi
 ## Settings
 
 :::details
-### What is Testnet?
-
-The testnet is an alternative Bitcoin blockchain, to be used for testing.
-Testnet coins (tBTC) are separate and distinct from actual bitcoins, and are never supposed to have any financial value.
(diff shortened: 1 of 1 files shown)


docs/FAQ/FAQ-UseWasabi.md
```diff
@@ -1255,25 +1255,6 @@ You can use the [Wasabi RPC server `gethistory` call](/using-wasabi/RPC.md#gethi
 
 ## Settings
```
MaxHillebrand, line 1257:
```suggestion

:::details
### How to switch to Testnet?

You can toggle between the Bitcoin `mainnet, testnet, or local regtest in the `Settings` tab.
This changes apply after a restart of Wasabi.

![](/SettingsNetwork.png)

Read more [here](/using-wasabi/Testnet.md).
:::
```

yahiheb: IMO this is not necessary.

MaxHillebrand: We also have something similar to the full node. 
A quick guide on the ways to activate it, and a link to the standalone chapter.
Especially in light of the search function, I think this is good to have.

yahiheb: I checked the full node FAQs and they add some value IMO, but for the testnet FAQ there is nothing that is not in the standalone chapter.

PR #754, opened 2020-04-13
Remove bech32 definition

yahiheb:
Anyone looking for an answer to that question should be familiar with bech32.

### DIFF
--- a/docs/FAQ/FAQ-UseWasabi.md
+++ b/docs/FAQ/FAQ-UseWasabi.md
@@ -334,8 +334,6 @@ This is a complete de-anonymization of your entire wallet!!
 :::details
 ### Why does Wasabi only use SegWit bech32 addresses?
 
-Wasabi generates bech32 addresses only, also known as `bc1q` addresses or native SegWit addresses.
-These addresses start with the characters `bc1q...`.
 Wasabi was created after the activation of SegWit, and it makes sense to support the most advanced address type, which has numerous benefits.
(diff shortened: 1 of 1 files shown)


docs/FAQ/FAQ-UseWasabi.md
```diff
@@ -334,8 +334,6 @@ This is a complete de-anonymization of your entire wallet!!
 :::details
 ### Why does Wasabi only use SegWit bech32 addresses?
 
-Wasabi generates bech32 addresses only, also known as `bc1q` addresses or native SegWit addresses.
-These addresses start with the characters `bc1q...`.
 Wasabi was created after the activation of SegWit, and it makes sense to support the most advanced address type, which has numerous benefits.
```
MaxHillebrand, line 337:
```suggestion
Wasabi generates bech32 addresses only, also known as `bc1q` addresses or native SegWit addresses.
Wasabi was created after the activation of SegWit, and it makes sense to support the most advanced address type, which has numerous benefits.
```

Agreed, the duplication is odd. 
But it might be worth considering to leave one of the references of `bc1q`?

yahiheb: But why should we do this?
we don't need introductions in the FAQs, IMO we need direct answers.

Look the question says ` Why does Wasabi only use SegWit bech32 addresses?`
and you are repeating that `Wasabi generates bech32 addresses only, also known as bc1q addresses or native SegWit addresses.`

PR #757, opened 2020-04-14
Remove false positive document

yahiheb:
That document is a mess right now with core and knots, I think we should just remove it.
If not then it should be carefully updated.


MaxHillebrand:
ACK - and further, I don't hear of users reporting this any more.

yahiheb:
A user just reported this on reddit https://www.reddit.com/r/WasabiWallet/comments/g1a4lh/what_should_this_represent/

PR #762, opened 2020-04-16
Suprise belt, shh secret!

molnard:
# Hereby

Let **Greg Hajdu** CEO of zkSNACKs to be promoted with a White belt for his brilliant finding on our upcoming CoinJoin technology naming: **Wabi Sabi!**

Image from iOS


molnard:
Let's merge it first.

PR #770, opened 2020-04-21
Update WasabiSetupTails.md

RiccardoMasutti:


PulpCattel:
Could be nice to also update the link in `Create admin password` to https://tails.boum.org/doc/first_steps/welcome_screen/administration_password/index.en.html since the current on is broken.

docs/using-wasabi/WasabiSetupTails.md
```diff
@@ -106,8 +106,13 @@ After the first time you save a Wasabi session, your persistent storage will loo
         |__ /Wallets
         |__ /Blocks
         |__ /BitcoinStore
+        |__ /Legal
 ```
 
+:::tip
+Make sure to copy `Legal` folder to avoid having to agree to Terms of Service and Privacy Policy every time you start Wasabi Wallet.
+:::
+
```
MaxHillebrand, line 115:
```suggestion
```

NACK - all the folders are important to keep, so why single out the legal folder as being important. It is much worse if users have do download filters and blocks on every start.

PR #794, opened 2020-04-30
fixing typo

Zolgarr:


Zolgarr:
Hm, I think both are correct, but I'd rather use `requesting payment`.

Zolgarr:
Let it be :D

PR #805, opened 2020-05-02
[USING] fix PayJoin

MaxHillebrand:
This ready for review branch adds some minor fixes to the PayJoin chapter, mainly that it works over clearnet too.
also adds a link to bip79 with a link to the btcpay spec for how our implementation differs.

### DIFF
--- a/docs/using-wasabi/BIPs.md
+++ b/docs/using-wasabi/BIPs.md
@@ -54,6 +54,15 @@ Here is a list of all the supported and integrated Bitcoin Improvement Proposals
 :::
 
 :::details
+### BIP 79: Bustapay: A Practical CoinJoin Protocol
+
+[BIP 79: Bustapay: A Practical CoinJoin Protocol](https://github.com/bitcoin/bips/blob/master/bip-0079.mediawiki)
+
+Wasabi deviates from BIP 79 in some aspects.
+The exact details of the implementation are [documented here](https://docs.btcpayserver.org/features/payjoin/payjoin-spec)
+:::
+
+:::details
 ### BIP 84: Derivation scheme for P2WPKH Based Accounts
 
 [BIP 84](https://github.com/bitcoin/bips/blob/master/bip-0084.mediawiki) defines a standard derivation scheme for hierarchical deterministic wallets BIP 32, specifically for segregated witness P2WPKH [BIP 173](BIPs.md#bip-173-base32-address-format-for-native-v0-16-witness-outputs).
--- a/docs/using-wasabi/PayJoin.md
+++ b/docs/using-wasabi/PayJoin.md
@@ -32,9 +32,10 @@ Further, it reduces the transaction fees paid by the merchant due to consolidati
 ## Coordination
 
 The coordination of this CoinJoin is done with the PayToEndPoint [P2EP] concept.
-The receiver opens a Tor hidden service, and accepts incoming connections to it.
+The receiver is reachable over the internet, either over a Tor hidden service or clearnet IP address, and accepts incoming connections to it.
 The link is included in a BIP21 Bitcoin URI, and is provided to the sender as the payment invoice.
-The sender uses this hidden service to connect to the server of the receiver and communicate the further protocol.
+The sender uses this hidden service or IP address to connect to the server of the receiver and communicate the further protocol.
+The coordination is usually done within seconds, and will abort to the fallback transaction after for example two minutes if the connection breaks.
 
 ## Fallback transaction
 
@@ -66,12 +67,12 @@ The sender broadcasts this transaction to the Bitcoin network.
 
 First and foremost, the common input ownership heuristic is broken, meaning that the inputs do not belong to one single entity, but to several of them.
 This breaks one of the most important assumptions of transaction surveillance companies.
-Contrarily to other CoinJoin implementations, the outputs are not of equal value, so it is not obvious that this transaction is in fact a CoinJoin.
+Contrarily to other CoinJoin implementations, the outputs are not of equal value, so it is not obvious that this transaction is in fact a CoinJoin, and that the assumption is wrong.
 
 Further, the outputs do not reflect the actual value of the transaction.
 In our example, the economic transfer is of `0.2 bitcoin` from Alice to Bob, but the outputs are worth `0.7 bitcoin` and `0.8 bitcoin`.
 This is obfuscating the actual amount payed.
 
-Finally, the receiver is consolidating his coins, which is a way for him to save on future transaction fees.
+Finally, the receiver is consolidating his coins, thus saving on future transaction fees.
 Without the PayJoin, the receiver would have two coins, worth `0.2 bitcoin` and `0.5 bitcoin`, and he would have to pay twice the transaction fee to spend these coins.
 But because of using PayJoin, he only has one coin, worth `0.7 bitcoin`, thus reducing his future transaction costs when spending this coin.
--- a/docs/using-wasabi/PayJoin.md
+++ b/docs/using-wasabi/PayJoin.md
@@ -34,10 +34,10 @@ Wasabi diverges from [the initial BIP 79 Bustapay proposal](https://github.com/b
 ## Coordination
 
 The coordination of this CoinJoin is done with the PayToEndPoint [P2EP] concept.
-The receiver is reachable over the internet, either over a Tor hidden service or clearnet IP address, and accepts incoming connections to it.
+The receiver is reachable over the internet, either over a Tor hidden service or clearnet IP address.
 The link is included in a BIP21 Bitcoin URI, and is provided to the sender as the payment invoice.
 The sender uses this hidden service or IP address to connect to the server of the receiver and communicate the further protocol.
-The coordination is usually done within seconds, and will abort to the fallback transaction after for example two minutes if the connection breaks.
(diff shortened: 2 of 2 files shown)


docs/using-wasabi/BIPs.md
```diff
@@ -53,6 +53,15 @@ Here is a list of all the supported and integrated Bitcoin Improvement Proposals
 [BIP 72: bitcoin: uri extensions for Payment Protocol](https://github.com/bitcoin/bips/blob/master/bip-0072.mediawiki)
 :::
 
+:::details
+### BIP 79: Bustapay: A Practical CoinJoin Protocol
+
+[BIP 79: Bustapay: A Practical CoinJoin Protocol](https://github.com/bitcoin/bips/blob/master/bip-0079.mediawiki)
+
+Wasabi deviates from BIP 79 in some aspects.
+The exact details of the implementation are [documented here](https://docs.btcpayserver.org/features/payjoin/payjoin-spec)
```
yahiheb, line 62: Does Wasabi support BIP79?

docs/using-wasabi/PayJoin.md
```diff
@@ -25,16 +25,19 @@
...
 The goal of the protocol is to break the common input ownership heuristic, while making it difficult to fingerprint that the transaction is in fact a CoinJoin.
 Further, it reduces the transaction fees paid by the merchant due to consolidation of coins.
 
+Wasabi diverges from [the initial BIP 79 Bustapay proposal](https://github.com/bitcoin/bips/blob/master/bip-0079.mediawiki), the details of the implemented specification are [documented here](https://docs.btcpayserver.org/features/payjoin/payjoin-spec).
+
 ## Coordination
 
 The coordination of this CoinJoin is done with the PayToEndPoint [P2EP] concept.
-The receiver opens a Tor hidden service, and accepts incoming connections to it.
+The receiver is reachable over the internet, either over a Tor hidden service or clearnet IP address, and accepts incoming connections to it.
```
yahiheb, line 37: This sentence is not clear to me, if we remove `either over a Tor hidden service or clearnet IP address` we end up with
`The receiver is reachable over the internet and accepts incoming connections to it.` which doesn't make sense.

docs/using-wasabi/PayJoin.md
```diff
@@ -25,16 +25,19 @@
...
 The link is included in a BIP21 Bitcoin URI, and is provided to the sender as the payment invoice.
-The sender uses this hidden service to connect to the server of the receiver and communicate the further protocol.
+The sender uses this hidden service or IP address to connect to the server of the receiver and communicate the further protocol.
+The coordination is usually done within seconds, and will abort to the fallback transaction after for example two minutes if the connection breaks.
```
yahiheb, line 40:
```suggestion
The coordination is usually done within seconds, and will abort to the fallback transaction after some time if the connection breaks.
```

docs/using-wasabi/PayJoin.md
```diff
@@ -66,12 +69,12 @@ The sender broadcasts this transaction to the Bitcoin network.
 
 First and foremost, the common input ownership heuristic is broken, meaning that the inputs do not belong to one single entity, but to several of them.
 This breaks one of the most important assumptions of transaction surveillance companies.
-Contrarily to other CoinJoin implementations, the outputs are not of equal value, so it is not obvious that this transaction is in fact a CoinJoin.
+Contrarily to other CoinJoin implementations, the outputs are not of equal value, so it is not obvious that this transaction is in fact a CoinJoin, and that the assumption is wrong.
```
yahiheb, line 72: I don't think that last part is necessary.
```suggestion
Contrarily to other CoinJoin implementations, the outputs are not of equal value, so it is not obvious that this transaction is in fact a CoinJoin.
```

docs/using-wasabi/BIPs.md
MaxHillebrand: mostly, yes.
There are some differences, mainly using PSBT instead of raw transactions. 
The couple deviations are in the btcpay server spec.

docs/using-wasabi/PayJoin.md
MaxHillebrand:
```suggestion
The receiver is reachable over the internet, either over a Tor hidden service or clearnet IP address.
```

docs/using-wasabi/BIPs.md
yahiheb: But Wasabi doesn't support receiver side does it?

MaxHillebrand: yes. But this doesn't mean that it is strictly compliant with BIP79.
For example, another thing with BIP79 is that the URI has a `bp=xxx` extension for `bustapay`.
Wasabi doesn't recognize this, I think, but only the BTCPay spec of `pj=xxx` for `PayJoin`.

yahiheb: Then maybe we shouldn't add this to the supported BIPs?

MaxHillebrand: For example, we also have BIP158 on there, but we don't 100% support it, cause we have segwit v0 single pubkey filters, not the basic filters. 

I think, that if we support the general idea / most of the details of a BIP, then we should reference it here.

BTCPay might open a new BIP with the updated specifics. If they do, then we should of course link to this one.

FYI, the changes really are minimal...
```
- The standard way of letting a sender know where to send a bustapay transaction is done via a bip21 encoded address. The key value "bpu" (short for "BustaPayUrl") should be used. An example of such address would be bitcoin:2NABbUr9yeRCp1oUCtVmgJF8HGRCo3ifpTT?bpu=https://bp.bustabit.com/submit It is highly encouraged that urls are kept short. 
+ The standard way of letting a sender know where to send a bustapay transaction is done via a bip21 encoded address. The key value "pj" (short for "payjoin URL") should be used. An example of such address would be bitcoin:2NABbUr9yeRCp1oUCtVmgJF8HGRCo3ifpTT?pj=https://example.com/submit It is highly encouraged that urls are kept short.

- Step 1. Sender creates a bitcoin transaction paying the receiver
+ Step 1. Sender creates a bitcoin transaction OR PSBT paying the receiver
+ While we also support the use of raw transaction instead of PSBT, we strongly advise against it.

- This transaction must use segwit for all inputs, and be fully valid and signed. The transaction must be eligible for propagation on the network (but not done so at this stage) 
+ If using PSBT, the PSBT must be finalized with witness UTXO information.
+ If using transactions, the transaction must use segwit for all inputs, and be fully valid and signed. The transaction must be eligible for propagation on the network (but not done so at this stage

- Step 2. Sender gives the "template transaction" to the receiver
+ Step 2. If using PSBT, sender gives "original PSBT" to the receiver, in base64 or hex format. 
+ If using transactions, sender gives "original transaction" to the receiver, in hex format.
- This is done via an HTTP POST request, sent to a "[bustapay] url" 
+ This is done via an HTTP POST request, sent to a "[payjoin] url" 

Step 3. Receiver processes the transaction and returns a partially signed coinjoin

- The receiver validates the transaction, and pays himself. The receiver then adds one or more of his own inputs (known as the contributed inputs) and (optionally) increases the output that pays himself (generally by the sum of the contributed inputs). Doing so creates a partial transaction, which the receiver returns to the sender. It is called such as it requires the sender to re-sign his own inputs. 
+ The receiver extracts the transaction, optionally verifying it is valid. (by calling testmempoolaccept on the extracted transaction)
+ The receiver extracts the feerate, adds one of his inputs and decreases the change output of the payer by the increased fee difference his input added such that the feerate is the same as the "orginal PSBT" or "original transaction".
```

yahiheb: Thank you for the explanation. I am just trying to make sure that we add what we really support.
I remember @nopara73 removing some bips before, so that is why I thought we shouldn't add it.

yahiheb: https://github.com/zkSNACKs/WalletWasabi/pull/3639#issue-415279379

Does this change anything?

MaxHillebrand: nope, I don't think so...

docs/using-wasabi/PayJoin.md
```diff
@@ -15,7 +15,7 @@
 
 1. Load a wallet and open the `Send` tab.
 
-2. Copy a [BIP21 Bitcoin URI](/using-wasabi/BIPs.md#bip-21-uri-scheme): with the flag `pj=` and paste it into the address field of the `Send` tab.
+2. Request from the receiver a [BIP21 Bitcoin URI](/using-wasabi/BIPs.md#bip-21-uri-scheme) with the flag `pj=` and paste it into the address field of the `Send` tab.
```
MaxHillebrand, line 18: I added this to make clear that the bip 21 comes from the receiver. A user asked this in the chat. 

@yahiheb could you please review, and then give a final merge ACK/NACK?

yahiheb, line 18: Maybe we should remove the flag part?
```suggestion
2. Request from the receiver a [BIP21 Bitcoin URI](/using-wasabi/BIPs.md#bip-21-uri-scheme) and paste it into the address field of the `Send` tab.
```

MaxHillebrand: well, if it's a BIP21 that does not have the pj flag, then it will just do a regular tx, no payjoin...

PR #808, opened 2020-05-02
Remove p2ep link

yahiheb:


docs/glossary/Glossary-PrivacyWasabi.md
```diff
@@ -156,7 +156,6 @@ Read more: [The importance of labeling](/using-wasabi/Receive.md#the-importance-
 ### Pay to EndPoint (P2EP)
 
 Pay to EndPoint is sending of bitcoins where the receiver adds one of his own coins as input for a two party CoinJoin.
```
MaxHillebrand, line 158:
```suggestion
Pay to EndPoint is when the receiver is reachable over the internet, via a Tor hidden service or IP address, and the sender communicates with the receiver to coordinate the transaction.
```

yahiheb: I don't think it is necessary to mention that the receiver is reachable over the internet ...

MaxHillebrand: but this is the entire point of it...

If you give me a regular Bitcoin address, then I cannot "reach you over the internet" to negotiate more advanced transactions. Contrarily, if you give me a P2EP BIP21 URI, then I have your endpoint, and thus can "reach you over the internet"

Maybe there's a better way of saying it other than "reach you over the internet"...

yahiheb: I get that, but I think in the glossary we are explaining terms in a very simple way.

Also a receiver might be `reachable over the internet, via a Tor hidden service or IP address` but that doesn't mean you can negotiate more advanced transactions.
I mean there is no mention of `P2EP BIP21 URI` in the suggestion.

MaxHillebrand: how about this?

```suggestion
Pay to EndPoint is when the receiver is reachable over the internet and the sender communicates with the receiver to coordinate a more advanced transaction.
The Tor hidden service, IP address or domain of the receiver is included in a BIP21 Bitcoin URI payment link.
```

Issue #813, opened 2020-05-07
Page stuck


BTCparadigm:
I think there is some problem in the Dojo page of the docs.
I can only move to pages that are under that section (Belt system, Wasabikas...) but nothing else. 

Nothing happens when I click on other titles or even the Wasabi logo.

This is reproducible and I tried it with 2 different computers and browsers (Win10 + Chrome & Firefox).

Go to docs.wasabiwallet.io -> Building Wasabi -> right click Dojo -> open in another page -> stuck.

Everything seems to work if I open Dojo with left click though.

I didn't find other pages in the docs that would have this problem.

yahiheb:
> Go to docs.wasabiwallet.io -> Building Wasabi -> right click Dojo -> open in another page -> stuck.

Tried that but it is working fine for me.

BTCparadigm:
Sorry for taking so long to reply. Here's a video clip from my view.
![2020-05-11-14-19-09](https://user-images.githubusercontent.com/51679301/81557618-f00c4400-9394-11ea-875c-1cbe27bc6aa7.gif)

yahiheb:
That is weird, It doesn't happen to me.

The only strange thing that I noticed is at the beginning of the video you are on the Dojo page https://docs.wasabiwallet.io/building-wasabi/Dojo.html but the link is https://docs.wasabiwallet.io/FAQ/

BTCparadigm:
Yeah, it's weird. The URL is correct until I go docs.wasabiwallet.io -> Building Wasabi -> right click Dojo -> open in another page. 
It opens the right page with correct URL, but after I try to move from Dojo page, the URL keeps changing but the page doesn't.

yahiheb:
> Yeah, it's weird. The URL is correct until I go docs.wasabiwallet.io -> Building Wasabi -> right click Dojo -> open in another page.
> It opens the right page with correct URL, but after I try to move from Dojo page, the URL keeps changing but the page doesn't.

I can reproduce it now, thanks.

MaxHillebrand:
maybe it's because of the `All Contributors` integration?

MaxHillebrand:
oh, the issue got automatically closed when #820 got merged... reopening until @BTCparadigm confirms it is fixed.

yahiheb:
I tested and it is not fixed (but we don't need the removed tag anyway)

> maybe it's because of the `All Contributors` integration?

[...]

PR #822, opened 2020-05-11
Change tag (Dojo.md)

yahiheb:
Second attempt to fix the issue #813


RiccardoMasutti:
?

MaxHillebrand:
Should we remove the All contributors in total?

yahiheb:
> Should we remove the All contributors in total?

I agree with that, it will fix the issue I think, and also that list is not up to-date.
I will remove it.

yahiheb:
Closing #823

PR #823, opened 2020-05-12
Remove all contributors

yahiheb:
That list is not up to-date and maybe some people don't want to be listed there.

Also is will maybe fix the issue #813


yahiheb:
> Is there anything of all contributors left in the repo?

I think there is nothing left.

PR #857, opened 2020-05-25
[USING] explain bip 39

MaxHillebrand:
This ready for review branch explains bip 39 mnemonic words

### DIFF
--- a/docs/using-wasabi/BIPs.md
+++ b/docs/using-wasabi/BIPs.md
@@ -38,7 +38,8 @@ Here is a list of all the supported and integrated Bitcoin Improvement Proposals
 :::details
 ### BIP 39: Mnemonic Code for Generating Deterministic Keys
 
-[BIP 39: Mnemonic Code for Generating Deterministic Keys](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki)
+[BIP 39: Mnemonic Code for Generating Deterministic Keys](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki) is a method to display the backup of private keys in easy to read sentence of 12, 18 or 24 words.
+It is based on a wordlist of 2048 words, where each word is unique in the first four characters for easy typo recognition.
 :::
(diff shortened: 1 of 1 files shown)


docs/using-wasabi/BIPs.md
```diff
@@ -38,7 +38,8 @@ Here is a list of all the supported and integrated Bitcoin Improvement Proposals
 :::details
 ### BIP 39: Mnemonic Code for Generating Deterministic Keys
 
-[BIP 39: Mnemonic Code for Generating Deterministic Keys](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki)
+[BIP 39: Mnemonic Code for Generating Deterministic Keys](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki) is a method to display the backup of private keys in easy to read sentence of 12, 18 or 24 words.
```
RiccardoMasutti, line 41:
```suggestion
[BIP 39: Mnemonic Code for Generating Deterministic Keys](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki) is a method to display the backup of private keys in easy to read sentence of 12, 18, 24 or more words.
```

MaxHillebrand: I don't think there are more than 24 words in a mnemonic standard?

RiccardoMasutti: You are right. I made this change because I don't want users to get confused with the seed + passphrase lenght.

MaxHillebrand: yeah, but the 25th word is another BIP, I believe?

lontivero, line 41: No. The bip39 is NOT a method to display the backup of private keys in easy to read sentence of 12, 18 or 24 words. It is a way to encode the seed used in the generation of deterministic wallets.

NACK. Why are we explaining the BIPs? They are really technical specifications.

PR #852, opened 2020-05-24
Correct Daemon commands for Windows

yahiheb:

PR #871, opened 2020-05-28
Edit daemon commands for macOS

yahiheb:


yahiheb:
> But You just edited this previously?

No I didn't edit this previously.
#852 corrected the Daemon commands for Windows
#872 added macOS' password finder commands

> Why the change back? I thought this works?

it turned out the `~` is not needed and people might copy even `~/Applications/Wasabi\ Wallet.app/Contents/MacOs$` when trying to execute the command.

This why it is more clear what to do and what to execute.

PR #874, opened 2020-05-28
Edit daemon commands for macOS when source is built

yahiheb:

PR #875, opened 2020-05-28
Edit daemon commands for Linux

yahiheb:
IMO this way it is more clear what should be done.

### DIFF
--- a/docs/using-wasabi/Daemon.md
+++ b/docs/using-wasabi/Daemon.md
@@ -37,14 +37,13 @@ A coin will be coinjoined into the first wallet until anonymity set target is re
 If the package is installed, execute in the command line in any directory: 
 
 ```bash
-~$ wassabee mix --wallet:MyFirstWallet --destination:MySecondWallet --keepalive
+wassabee mix --wallet:MyFirstWallet --destination:MySecondWallet --keepalive
 ```
 
-If the source code is built:
+If the source code is built, open the Command Line and navigate to the WalletWasabi.Gui folder (inside the cloned repo) and execute:
 
 ```bash
-~$ cd ~/WalletWasabi/WalletWasabi.Gui
-~/WalletWasabi/WalletWasabi.Gui$ dotnet run -- mix --wallet:MyFirstWallet --destination:MySecondWallet --keepalive
+dotnet run -- mix --wallet:MyFirstWallet --destination:MySecondWallet --keepalive
 ```
 
 ### macOS
--- a/docs/using-wasabi/Daemon.md
+++ b/docs/using-wasabi/Daemon.md
@@ -40,9 +40,10 @@ If the package is installed, execute in the command line in any directory:
 wassabee mix --wallet:MyFirstWallet --destination:MySecondWallet --keepalive
 ```
 
-If the source code is built, open the Command Line and navigate to the WalletWasabi.Gui folder (inside the cloned repo) and execute:
+If the source code is built, open the command line terminal and execute:
 
 ```bash
+cd ~/WalletWasabi/WalletWasabi.Gui
 dotnet run -- mix --wallet:MyFirstWallet --destination:MySecondWallet --keepalive
 ```
 
--- a/docs/using-wasabi/Daemon.md
+++ b/docs/using-wasabi/Daemon.md
@@ -40,10 +40,9 @@ If the package is installed, open the command line terminal and execute in any d
 wassabee mix --wallet:MyFirstWallet --destination:MySecondWallet --keepalive
 ```
 
-If the source code is built, open the command line terminal and execute:
+If the source code is built, open the terminal, navigate to the `WalletWasabi.Gui` folder inside the cloned repository, and execute:
 
 ```bash
-cd ~/WalletWasabi/WalletWasabi.Gui
 dotnet run -- mix --wallet:MyFirstWallet --destination:MySecondWallet --keepalive
 ```
 
--- a/docs/using-wasabi/Daemon.md
+++ b/docs/using-wasabi/Daemon.md
@@ -34,7 +34,7 @@ A coin will be coinjoined into the first wallet until anonymity set target is re
 
 ### Linux
 
-If the package is installed, execute in the command line in any directory: 
+If the package is installed, open the command line terminal and execute in any directory:
 
 ```bash
 wassabee mix --wallet:MyFirstWallet --destination:MySecondWallet --keepalive


MaxHillebrand:
from the feedback I get from n00b users, often times they don't know how to `navigate` through folders in command line. 
I would rather have true step by step guide that any one can easily follow, rather than assuming that this is understood.

so, NACK

yahiheb:
> from the feedback I get from n00b users, often times they don't know how to navigate through folders in command line.

That is easy to solve by just a looking on the internet or by asking.

> I would rather have true step by step guide that any one can easily follow, rather than assuming that this is understood.

The proposed change is a step by step guide.

The problem with the current document is that `n00b users` will copy the text as is and execute it which will not work.

For example the following is not a command:
`~/WalletWasabi/WalletWasabi.Gui$ dotnet run -- mix --wallet:MyFirstWallet --destination:MySecondWallet --keepalive`

So for `n00b users` the `~/WalletWasabi/WalletWasabi.Gui$` part should definitely be removed.

docs/using-wasabi/Daemon.md
```diff
@@ -37,14 +37,13 @@ A coin will be coinjoined into the first wallet until anonymity set target is re
 If the package is installed, execute in the command line in any directory: 
 
 ```bash
-~$ wassabee mix --wallet:MyFirstWallet --destination:MySecondWallet --keepalive
+wassabee mix --wallet:MyFirstWallet --destination:MySecondWallet --keepalive
 ```
 
-If the source code is built:
+If the source code is built, open the Command Line and navigate to the WalletWasabi.Gui folder (inside the cloned repo) and execute:
```
MaxHillebrand, line 43:
```suggestion
If the source code is built, open the command line terminal and execute:
```

docs/using-wasabi/Daemon.md
```diff
@@ -37,14 +37,13 @@ A coin will be coinjoined into the first wallet until anonymity set target is re
...
 
 ```bash
-~$ cd ~/WalletWasabi/WalletWasabi.Gui
-~/WalletWasabi/WalletWasabi.Gui$ dotnet run -- mix --wallet:MyFirstWallet --destination:MySecondWallet --keepalive
+dotnet run -- mix --wallet:MyFirstWallet --destination:MySecondWallet --keepalive
```
MaxHillebrand, line 46:
```suggestion
cd ~/WalletWasabi/WalletWasabi.Gui
dotnet run -- mix --wallet:MyFirstWallet --destination:MySecondWallet --keepalive
```

yahiheb: If you open the Terminal anywhere and you execute the first command, will it the directory change to `WalletWasabi.Gui`  folder no matter where it is?

If yes then what happens when you have two repos? https://github.com/zkSNACKs/WasabiDoc/pull/874#discussion_r432231047

MaxHillebrand: Yes. the `~/` denotes the root directory. Thus, if the user did `~$ git clone github.com/zksnacks/WalletWasabi`, as is earlier in the guide, then the Gui folder will be at that path.

Sure, if there are two repos `WalletWasabi1` and `WalletWasabi2`, then this would not work. But, this is only a small sub section of users who will clone the repo twice, as it is not mentioned in the docs to do so. And whoever clones the twice probably knows how to change directories anyhow...

yahiheb: > Yes. the ~/ denotes the root directory. Thus, if the user did ~$ git clone github.com/zksnacks/WalletWasabi, as is earlier in the guide, then the Gui folder will be at that path.

> Sure, if there are two repos WalletWasabi1 and WalletWasabi2, then this would not work.

Not only that, even if you have only one Wasabi repo but it is not in the root directory then it will not work.

My point is using `cd ~/WalletWasabi/WalletWasabi.Gui` will only work assuming that the repo is somewhere, while saying `open the Command Line and navigate to the WalletWasabi.Gui folder (inside the cloned repo)` doesn't assume anything.
And if the user is not able to navigate to a folder or is not willing to learn how to do that, how do you expect them to clone the repo and use the daemon?

MaxHillebrand: That is a good point. 
But, if the user does a step by step according to the docs, the GUI folder will be on this path specifically. Sure, when he deviates from the guide, he cannot expect that it still works.

I'm kinda torn here... I know several whom I helped to build source, and they struggle with basic commands like `cd` already... Thus, I'd like this guide to be fool proof. I'd prefer it to show every [reasonable] action that the user has to take, and this included `cd`.

MaxHillebrand:
We can go back and forth on it for ever, so I'd like to make a decision on this...

- the documentation should have a step by step guide that a new user can follow. [i.e. cd command should be included]
- the guide must be in it of itself correct, but if the user deviates form the guide, there's no guarantee that the following commands still work [i.e. cloning repo in a different folder]

I will now commit the suggestions so that they adhere to the above statements.

Then we can go over another round of review and N/ACK.

docs/using-wasabi/Daemon.md
```diff
@@ -37,14 +37,14 @@ A coin will be coinjoined into the first wallet until anonymity set target is re
 If the package is installed, execute in the command line in any directory:
```
MaxHillebrand, line 37:
```suggestion
If the package is installed, open the command line terminal and execute in any directory:
```

yahiheb:
> We can go back and forth on it for ever, so I'd like to make a decision on this...
> 
> * the documentation should have a step by step guide that a new user can follow. [i.e. cd command should be included]
> * the guide must be in it of itself correct, but if the user deviates form the guide, there's no guarantee that the following commands still work [i.e. cloning repo in a different folder]
> 
> I will now commit the suggestions so that they adhere to the above statements.
> 
> Then we can go over another round of review and N/ACK.

If we follow this rational we then should make guides how to open a terminal, in case someone doesn't know how to do that.
I think we are treating our users as not kids or stupid (specially linux users who clone repos and try to use the daemon).

Again if saying `If the source code is built, open the Command Line and navigate to the WalletWasabi.Gui folder (inside the cloned repo) and execute:` is not clear in and of itself then the user is better not to touch the daemon.

If someone was able to clone a repository we should expect them to be able to navigate to that folder.

MaxHillebrand:
> If someone was able to clone a repository we should expect them to be able to navigate to that folder.

The thing is, that I've encountered at least three users in recent memory who did not know how to change directory... 

Sure, we don't need to tell them how to `turn on the pc`, `put your hands on the keyboard` and `open a terminal`, but, at least `what to do in the terminal` should be explained.

docs/using-wasabi/Daemon.md
```diff
@@ -34,17 +34,17 @@ A coin will be coinjoined into the first wallet until anonymity set target is re
...
-If the package is installed, execute in the command line in any directory: 
+If the package is installed, open the command line terminal and execute in any directory:
 
 ```bash
-~$ wassabee mix --wallet:MyFirstWallet --destination:MySecondWallet --keepalive
+wassabee mix --wallet:MyFirstWallet --destination:MySecondWallet --keepalive
 ```
 
-If the source code is built:
+If the source code is built, open the command line terminal and execute:
```
MaxHillebrand, line 43:
```suggestion
If the source code is built, open the terminal, navigate to the `WalletWasabi.Gui` folder inside the cloned repository, and execute:
```

docs/using-wasabi/Daemon.md
```diff
@@ -34,17 +34,17 @@ A coin will be coinjoined into the first wallet until anonymity set target is re
...
 
 ```bash
-~$ cd ~/WalletWasabi/WalletWasabi.Gui
-~/WalletWasabi/WalletWasabi.Gui$ dotnet run -- mix --wallet:MyFirstWallet --destination:MySecondWallet --keepalive
+cd ~/WalletWasabi/WalletWasabi.Gui
```
MaxHillebrand, line 46:
```suggestion
```

PR #878, opened 2020-06-02
Clickable elements too close together

RiccardoMasutti:
Touch elements are so close to each other that a mobile user cannot easily tap a desired element with their finger without also tapping a neighboring element.

This was reported by Google Smartphone crawler and this PR should help overall SEO and mobile usability:
Wasabi-docs-gsc

### DIFF
--- a/docs/.vuepress/styles/index.styl
+++ b/docs/.vuepress/styles/index.styl
@@ -56,7 +56,7 @@ body
     font-weight 700
 
   details
-    margin -4rem 0 0 0
+    margin -3.5rem 0 0 0
     &[open]
       > :last-child
         margin-bottom 3rem


docs/.vuepress/styles/index.styl
```diff
@@ -56,7 +56,7 @@ body
     font-weight 700
 
   details
-    margin -4rem 0 0 0
+    margin -3.5rem 0 0 0
```
yahiheb, line 59: Just to understand what does -3.5rem mean? and does it solve the reported issue?

RiccardoMasutti: -3.5 pixels of margin property of the root element. The `rem` units refer to the property’s initial value.

It solves the issue but I need to apply the edit and verify via GSC/Google mobile spider
fix mobile

MaxHillebrand:
Ok, I'll merge this, then please verify that it is fixed now @RiccardoMasutti.

PR #881, opened 2020-06-07
[USING] Daemon make clear with white space

MaxHillebrand:
This works on Linux, not sure if it also works on Mac and Windows, can someone please test?

### DIFF
--- a/docs/using-wasabi/Daemon.md
+++ b/docs/using-wasabi/Daemon.md
@@ -21,6 +21,7 @@ This daemon is especially useful for power users mixing bitcoin in the backend o
 `mix` makes sure Wasabi starts in daemon and not the GUI.
 
 `--wallet:` specifies the name of the hot wallet with the coins you want to CoinJoin.
+Capitalization and whitespace is important, so for example `--wallet:"Hot Wallet".
 
 `--destination:` specifies the destination wallet that the mixed coins will be coinjoined into, after the target anonymity set is reached.
 A coin will be coinjoined into the first wallet until anonymity set target is reached, then there will be one additional CoinJoin into the `destination` wallet.


docs/using-wasabi/Daemon.md
```diff
@@ -21,6 +21,7 @@ This daemon is especially useful for power users mixing bitcoin in the backend o
 `mix` makes sure Wasabi starts in daemon and not the GUI.
 
 `--wallet:` specifies the name of the hot wallet with the coins you want to CoinJoin.
+Capitalization and whitespace is important, so for example `--wallet:"Hot Wallet".
```
yahiheb, line 24: Because this is valid for the destination address as well, we should add this under the `--destination:` section.
```suggestion
```

docs/using-wasabi/Daemon.md
```diff
@@ -21,6 +21,7 @@ This daemon is especially useful for power users mixing bitcoin in the backend o
...
 
 `--destination:` specifies the destination wallet that the mixed coins will be coinjoined into, after the target anonymity set is reached.
 A coin will be coinjoined into the first wallet until anonymity set target is reached, then there will be one additional CoinJoin into the `destination` wallet.
```
yahiheb, line 27:
```suggestion
A coin will be coinjoined into the first wallet until anonymity set target is reached, then there will be one additional CoinJoin into the `destination` wallet.

Capitalization is important when specifying the name of the wallet, and if it contains a white-space you should use the double quotation marks like this: `--wallet:"Hot Wallet"`
```

docs/using-wasabi/Daemon.md
lontivero, line 24: File names case sensitivity and how to deal with white spaces are shell dependent. Windows is case insensitive while in linux and mac are case sentitive but it also depends on the shell, in bash (and other POSIX compatible shells) you write `Hello\ World` while in fish for example you write `'Hello Word'` (or "Hello World"). This is also valid for Windows because it is not the same CMD, 4DOS (i think nobody use it anymore), Powershell, etc.

I have seen people that doesn't know what a log file is, what a password is for and why they cannot spend their bitcoin cash with Wasabi. One guy told me today that he doesn't know that is the "path"! 
Documenting these kind of things are completely out of scope.

MaxHillebrand: Thanks @lontivero, I didn't know that it was this inconsistent across OS... 

Then I agree, it should not be part of the docs...

PR #641, opened 2020-03-16
Add structured data to FAQs

RiccardoMasutti:
Follows https://github.com/zkSNACKs/WalletWasabi/pull/2326

Marking up content on Wasabi Documentation website can:
- Lead to the generation of rich snippets in search engine results like the one showed down here
- Enhance CTR (Click Through Rate) from the search result
- Provide greater information to search engines to improve their understanding of the content on Wasabi website

As specified on [https://developers.google.com/search/docs/data-types/faqpage](https://developers.google.com/search/docs/data-types/faqpage)

Test code: https://search.google.com/structured-data/testing-tool

Final result:
image

PR #882, opened 2020-06-09
Add structured data to FAQs

RiccardoMasutti:
This will boost SEO for docs.wasabiwallet.io
Replace #641

Marking up content on Wasabi Documentation website can:
- Lead to the generation of rich snippets in search engine results like the one showed down here
- Enhance CTR (Click Through Rate) from the search result
- Provide greater information to search engines to improve their understanding of the content on Wasabi website

As specified on [https://developers.google.com/search/docs/data-types/faqpage](https://developers.google.com/search/docs/data-types/faqpage)

Test code: https://search.google.com/structured-data/testing-tool

Final result:
image

### DIFF
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -50,9 +50,9 @@ module.exports = {
             title = ''
           }
           const slug = slugify(title)
-          return `<details id="${slug}"><summary><a href="#${slug}" aria-hidden="true" class="header-anchor">#</a> <h4>${title}</h4></summary>`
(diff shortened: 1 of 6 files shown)


RiccardoMasutti:
@dennisreimann unfortunately this also applies to the `Glossary`.
Imho it's not wrong to use FAQ markups for glossary, but would be better to use `DefinedTerm` as it's (https://schema.org/DefinedTerm):

> A word, name, acronym, phrase, etc. with a formal definition. Often used in the context of category or subject classification, glossaries or dictionaries, product or creative work types, etc.

Would be awesome to create a custom layout for that specific page.
Example:

```
  <div itemscope itemtype="https://schema.org/DefinedTerm">
    <dt itemprop="name"><dfn><abbr>Fungibility</abbr></dfn></dt>
    <dd itemprop="description">Fungibility is a desirable property of Bitcoin UTXOs that are indistinguishable from each other.</dd>
  </div>
```

MaxHillebrand:
I still don't understand why you closed #641 and why now #882 is better?

MarnixCroes (commented):
Do we still want to have this?
Are these items still correct?

MarnixCroes:
talked with @CAnorbo , he knows about SEO.
But wouldn't approach it this way.
closing

PR #884, opened 2020-06-14
Add WabiSabi

RiccardoMasutti:
Adds WabiSabi under `Research` and change menu order


yahiheb:
NACK. IMO we shouldn't add this.
If we want to add anything about WabiSabi we can link to the repository or to the gist https://gist.github.com/nothingmuch/9427a27e29ab3c525a23ae4fd6f8a5ae

PR #886, opened 2020-06-16
Fix Release name

nopara73:
Suggested by https://twitter.com/CoinJoinDay

### DIFF
--- a/docs/using-wasabi/InstallPackage.md
+++ b/docs/using-wasabi/InstallPackage.md
@@ -97,7 +97,7 @@ Verify that the fingerprint is `${zksnacksPublicKeyFingerprint}`.
 
 ![](/DownloadTar.png)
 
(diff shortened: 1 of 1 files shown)


yahiheb:
Why only for the other Linux distributions the name of the package and the signature file are like `WasabiLinux.Version.extension` while for windows, macOS, Debian/Ubuntu they are like `Wasabi.Version.extension`?

nopara73:
If it has a reason you can find it here: https://github.com/zkSNACKs/WalletWasabi/blob/master/WalletWasabi.Packager/Program.cs

nopara73:
Well, this is the correct name. Now @yahiheb did a PR that changes this https://github.com/zkSNACKs/WalletWasabi/pull/3780, so it's kind of up to you how you want to handle it, there are multiple good ways to proceed forward.

PR #887, opened 2020-06-17
Update AddressReuse.md

RiccardoMasutti:


lontivero:
In fact, the dust attack can be somehow effective in wallets that doesn't have coin control. In Wasabi this is much less effective and we have that dust threshold only to remove the temptation of some users to consolidate dust coins with other coins (most of the times those dust coins cannot be spent along)

docs/why-wasabi/AddressReuse.md
```diff
@@ -57,4 +57,4 @@ BIP 44 HD wallets use an alternative key derivation function called "hardened de
 When given a hardened child private key, then the parent private key cannot be calculated.
 An address gets removed from the GUI in the `Receive` tab as soon as a coin is received to it.
 
-To protect against forced address reuse attack (Dusting), Wasabi has a modifiable dust limit, where the wallet does not show coins below a certain threshold value.
+To protect against forced address reuse attack (Dusting), Wasabi has a [modifiable dust limit](/FAQ/FAQ-UseWasabi.md#what-is-the-dust-threshold), where the wallet does not show coins below a certain threshold value.
```
yahiheb, line 60:
```suggestion
To protect against forced address reuse attack (Dusting), Wasabi has a modifiable [dust limit](/FAQ/FAQ-UseWasabi.md#what-is-the-dust-threshold), where the wallet does not show coins below a certain threshold value.
```

PR #906, opened 2020-07-13
Update how to activate/deactivate LWM

yahiheb:


yahiheb:
> how about `toggle` instead of `activate/deactivate`?

I think `activate/deactivate` is fine.

PR #925, opened 2020-08-24
Remove riot.im from supported or chat channels

yahiheb:
This PR removes Riot as it is not officially supported and I don't think anyone uses that Riot Wasabi room.


RiccardoMasutti (approved):
ACK

nopara73:
@yahiheb could you remove the rest except reddit? We don't give support anywhere else.

MaxHillebrand:
late ACK

PR #926, opened 2020-08-25
Remove non official stuff

yahiheb:
There is one place left where Telegram and keybase are mentioned as chat groups, should those be removed as well?
https://docs.wasabiwallet.io/FAQ/FAQ-Introduction.html#where-can-i-find-wasabi-wallet-on-social-media


MaxHillebrand:
late ACK.

PR #944, opened 2020-09-17
Update node-forge

nopara73:


nopara73:
That's not how you update it seems.

molnard:
package.lock is a generated file. It extracts all the used references recursively. Probably it is from NBitcoin? If yes it should be updated there.

nopara73:
image

molnard:
Nevermind.

PR #946, opened 2020-09-21
[FAQ] how to change coordinator hidden service

MaxHillebrand:
This adds a quick FAQ about how to change the coordinator onion.

### DIFF
--- a/docs/FAQ/FAQ-UseWasabi.md
+++ b/docs/FAQ/FAQ-UseWasabi.md
@@ -1412,7 +1412,7 @@ Please note that Wasabi is designed for the dark theme, and some color schemes m
 ### How can I connect to a different coordinator?
 
 To reach the coordinator you need their hidden service .onion address.
-Open the `Conf.json` file through the top menu bar.
+Open the `Config.json` file through the top menu bar.
 
 ![Wasabi Wallet File Options](/MenuFileOpen.png "Wasabi Wallet File Options")
 
--- a/docs/FAQ/FAQ-UseWasabi.md
+++ b/docs/FAQ/FAQ-UseWasabi.md
@@ -1412,7 +1412,7 @@ Please note that Wasabi is designed for the dark theme, and some color schemes m
 ### How can I connect to a different coordinator?
 
 To reach the coordinator you need their hidden service .onion address.
-Open the `Conf.json`  file through the top menu bar.
+Open the `Conf.json` file through the top menu bar.
 
 ![Wasabi Wallet File Options](/MenuFileOpen.png "Wasabi Wallet File Options")
 
--- a/docs/FAQ/FAQ-UseWasabi.md
+++ b/docs/FAQ/FAQ-UseWasabi.md
@@ -1408,6 +1408,17 @@ Please note that Wasabi is designed for the dark theme, and some color schemes m
 ![Wasabi Wallet white theme](/WhiteMode.png "Wasabi Wallet white theme")
 :::
 
+:::details
+### How can I connect to a different coordinator?
+
+To reach the coordinator you need their hidden service .onion address.
+Open the `Conf.json`  file through the top menu bar.
+
+![Wasabi Wallet File Options](/MenuFileOpen.png "Wasabi Wallet File Options")
+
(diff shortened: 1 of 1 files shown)


docs/FAQ/FAQ-UseWasabi.md
```diff
@@ -1408,6 +1408,17 @@ Please note that Wasabi is designed for the dark theme, and some color schemes m
 ![Wasabi Wallet white theme](/WhiteMode.png "Wasabi Wallet white theme")
 :::
 
+:::details
+### How can I connect to a different coordinator?
+
+To reach the coordinator you need their hidden service .onion address.
+Open the `Conf.json`  file through the top menu bar.
```
RiccardoMasutti, line 1415:
```suggestion
Open the `Conf.json` file through the top menu bar.
```

RiccardoMasutti (changes requested):

docs/FAQ/FAQ-UseWasabi.md
```diff
@@ -1408,6 +1408,17 @@ Please note that Wasabi is designed for the dark theme, and some color schemes m
 ![Wasabi Wallet white theme](/WhiteMode.png "Wasabi Wallet white theme")
 :::
 
+:::details
+### How can I connect to a different coordinator?
+
+To reach the coordinator you need their hidden service .onion address.
+Open the `Conf.json` file through the top menu bar.
```
DanGould, line 1415: change line to 2. From menu: File/Open/Config File.

`Conf.json` -> `Config.json`

MaxHillebrand:
```suggestion
Open the `Config.json` file through the top menu bar.
```

yahiheb:
Why should we add this to the docs?

DanGould:
@yahiheb because it answers a question that is asked frequently

MaxHillebrand:
@yahiheb because Wasabi =/ zkSNACKs.

although this repository ships a package that by default connects to the zkSNACKs coordinator, there are already and will continue to be other coordinators operated. That's a feature, not a bug!!

 Users can and will connect to different coordinators, so it ought to be explained how to do it properly, so to not mess up.

RiccardoMasutti:
> @yahiheb because Wasabi =/ zkSNACKs.
> 
> although this repository ships a package that by default connects to the zkSNACKs coordinator, there are already and will continue to be other coordinators operated. That's a feature, not a bug!!
> 
> Users can and will connect to different coordinators, so it ought to be explained how to do it properly, so to not mess up.

I agree on explaining how to change the coordinator variable, but I believe that the lack of liquidity in the current other coordinators can be a privacy issue and I would advise against most users to change it.

MaxHillebrand:
yes Riccardo, this is a concern, and notice that's the reason why I do not link to the onion of a different coordinator, as this would be a recommendation, and thus carefully ought to be considered.

Nevertheless, changing the coordinator onion, even if it's for your own setup on regtest, is useful information worthy to be in the docs.

yahiheb:
Yes Wasabi =! zkSNACKs, but the docs are about Wasabi stuff mainly, so why should we add anything not really related to Wasabi?
I mean does Wasabi have any other coordinator that users should manually connect to? AFAIK no, so there is no need for this addition.

> @yahiheb because it answers a question that is asked frequently

But that question is not related to Wasabi.
Don't get me wrong I am all in for chaincase but I don't think we should add this here just because people want to use Wasabi to connect to chaincase's coordinator.

MaxHillebrand:
> why should we add anything not really related to Wasabi?

Because changing the coordinator is a feature in the Wasabi client.

This is not the zkSNACKs documentation [although it's of course in the zkSNACKs repo, lol], but the Wasabi documentation. And the Wasabi client, is designed in such a way to connect to different than the default coordinator.

imho such a core architecture feature ought to be documented, specifically when there are real use cases for changing it.

> I mean does Wasabi have any other coordinator that users should manually connect to?

Yes, for example the mainchain coordinator of chaincase, and I know several users who connect to it with their Wasabi desktop client.

I think it's a reasonable assumption to see in the future new coordinators with unique round parameters.

RiccardoMasutti:
@MaxHillebrand @yahiheb What if we add a `:::warning` explaining that by changing the default coordinator, the user expose himself to a series of problems not related to the default coordinator? (eg. low liquidity, less privacy/anonset per round, etc.)

PR #947, opened 2020-09-21
Add Bitcoin Core variable

RiccardoMasutti:


MaxHillebrand (changes requested):
we're packaging knots binaries, but support of core binaries too...

though I think, `bitcoinKnotsVersion` is more accurate.

yahiheb:
NACK.
Those version numbers are only used in the tails document as an example so there is no need for a special variable for them, we don't need to maintain this.

MaxHillebrand:
That's a good point @yahiheb.

Then the question is, why do we keep the version numbers in this tails chapter? Can we rewrite the text so that this is not mentioned, thus reducing the maintenance burden further?

yahiheb:
Those are just examples, I think it is ok to keep them as they are IMO.

yahiheb:
@RiccardoMasutti Don't open branches in the repo directly.

MaxHillebrand:
well, if there are the version numbers mentioned, and Core upgrades every 6 months [Knots even more frequent, I think], then I think this justifies a version number so to reduce maintenance efforts in the future.

yahiheb:
We don't need to change those version numbers, they are just examples.

MaxHillebrand (changes requested):
I see @yahiheb, now I understand your rational, and I agree.

NACK 5ba85c4

RiccardoMasutti:
Ok, let's close it

PR #954, opened 2020-11-16
Upgrade dotnet version to .NET 5.0

yahiheb:
Should only be merged at least after https://github.com/zkSNACKs/WalletWasabi/pull/4730 is merged. (maybe only after the next release)


RiccardoMasutti (approved):
ACK.

MaxHillebrand (approved):
tACK.

https://github.com/zkSNACKs/WalletWasabi/pull/4730 is merged now, meaning that this 'is released' for those who build source. I think this should be merged now.

PR #965, opened 2021-01-20
Remove Dojo

RiccardoMasutti:


MaxHillebrand (approved):
I still like the idea, but unfortunately it wasn't really used often.
Maybe we can revive it with the karma in GitHub discussions?

RiccardoMasutti:
> I still like the idea, but unfortunately it wasn't really used often.
> Maybe we can revive it with the karma in GitHub discussions?

We reached a consensus and decided to remove the Dojo from the documentation.
The final goal will be to move everything to GitHub discussions

PR #979, opened 2021-04-06
Remove Tails setup

BTCparadigm:
This removes Tails markdown files and navigation bar links, because Tails is not supported an it has high maintenance burden.


MaxHillebrand (approved):
ACK 33d33d6

yahiheb:
As agreed this shouldn't be removed, but rather the document should be corrected.

PR #988, opened 2021-05-07
Add broadcast RPC call

RiccardoMasutti:
Follows https://github.com/zkSNACKs/WalletWasabi/pull/5436

WIP, needs testing and review


lontivero:
We should test it but it doesn't look correct to me because it pass an object without name when it should pass simply the hex. Does it work for you? If not then use an array `"params": ["the hex goes here"]`

MarnixCroes:
This is part of the RPC docs now.

PR #1010, opened 2021-07-29
fix selectwallet example

KenKundert:
I found that I needed to add `"id:"1",` to get these examples to work.


RiccardoMasutti (approved):

yahiheb (approved):
utACK

PR #1015, opened 2021-10-17
Replace references to the Gui project by Fluent.Desktop project

yahiheb:
Following the removal of the Gui project https://github.com/zkSNACKs/WalletWasabi/pull/6492

PR #1016, opened 2021-10-18
Remove any command line instructions to run as a daemon

yahiheb:
https://github.com/zkSNACKs/WasabiDoc/pull/1015#issuecomment-945227525


molnard:
We have to be very conscious here. docs.wasabiwallet.io is an archive and needs to be kept for WW1 as it is. Basically, this documentation belongs to the backport branch here. WW2 documentation will be on another subdomain.

molnard:
The links there might need to change as we keep removing stuff from the master, but the backport branch will hold every WW1-related stuff until the end of the time.

PR #1017, opened 2021-10-24
Clarifying key verification on Linux

BTCparadigm:
I remember finding the instructions a bit confusing when I first tried it and I've heard the same feedback a few times, once just yesterday. This PR is supposed to clarify the instructions.

1. There's no option to download the key file
2. Making sure every relevant line is included
3. Application suggestion
4. Other clarifications

Oh yeah, I accidentaly updated the file back to the original one after making changes, and then reverted it.

### DIFF
--- a/docs/using-wasabi/InstallPackage.md
+++ b/docs/using-wasabi/InstallPackage.md
@@ -62,7 +62,7 @@ Among others, here is where your wallet files and your logs reside.
 
 If you have already imported zkSNACKs' PGP public key, then jump to step 2.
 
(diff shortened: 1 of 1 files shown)


lontivero:
PGP verification is necessary only for linux users that do not run it from code. We are now targeting a broader user base that will not be linux users by sure. But even in the case of highly technical users that know what they do, how many of them do you estimate can verify the zkSNARCKs PGP key is the correct one? I believe the number is close to 0%.

PR #1018, opened 2021-11-02
Bump client version to 1.1.13

yahiheb:
https://github.com/zkSNACKs/WalletWasabi/pull/6609

### DIFF
--- a/docs/.vuepress/variables.js
+++ b/docs/.vuepress/variables.js
@@ -1,6 +1,6 @@
 // Add custom variables here:
 const variables = {
-  currentVersion: '1.1.12.9',
+  currentVersion: '1.1.13',
   dotnetVersion: '5.0',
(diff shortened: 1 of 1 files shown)


lontivero:
I think this is not right because `master` doesn't have `1.1.13`. I mean, if someone want to have that specific version it needs to switch to `backport` and compile branch. Right now master is more like a `v2.0.0.0-alpha-preview-very-soon-stay-tuned` version.

yahiheb:
> I think this is not right because `master` doesn't have `1.1.13`. I mean, if someone want to have that specific version it needs to switch to `backport` and compile branch. Right now master is more like a `v2.0.0.0-alpha-preview-very-soon-stay-tuned` version.

Yes but the current documentation is for WW1.0

lontivero:
ohhh sorry, you're right.

lontivero (approved):

PR #1079, opened 2022-06-14
Update Download pics

BTCparadigm:
Add 2 gitignores.
Remove non-existing youtube link.
Update pics.

Part of updating `InstallPackage.md`.
Untested.


.gitignore
```diff
@@ -3,3 +3,5 @@
 /.vs/ProjectSettings.json
 /.vs/WasabiDoc/config/applicationhost.config
 /.vs/slnx.sqlite
+/.vs/slnx.sqlite-journal
+/.vs/WasabiDoc/v16/.suo
```
yahiheb, line 7: What are those? Why should we ignore them?

BTCparadigm: Some stuff that gets added to my vs automatocally when I open docs. I can put them to my local gitignore if thats better?

PR #1090, opened 2022-07-10
[Docs-2.0] Update `WalletLoad.md`

BTCparadigm:
Removed 3 old pics. 
Added 2 new ones instead. Not perfect but should be ok for now.
Updated content to 2.0.


yahiheb (changes requested):
`WalletManagerLoadWallet.png` should also be removed and updated in https://github.com/zkSNACKs/WasabiDoc/blob/docs-2.0/docs/building-wasabi/TechnicalOverview.md

Issue #1022, opened 2021-12-22
Plan for docs 2.0


BTCparadigm:
Build in a different [branch](https://github.com/zkSNACKs/WasabiDoc/tree/docs-2.0) without changing categories or breaking internal or external links too much.
NOTE: No need to say 2.0 in new docs.

Update all content except WHY WASABI and BUILDING WASABI.

## USING WASABI

Done:
- [x] 10Commandments #1050
- [x] BackupBestPractices
- [x] BitcoinFullNode #1107 
- [x] ELI5 #1049 
- [x] IndustryStandards #1092 
- [x] LockScreen #1092
- [x] PasswordBestPractices
- [x] PrivacyMode #1086 & #1089
- [x] WalletLoad #1090 
- [x] Testnet #1091
- [x] RestoreElectrum #1105 

ToDo:
- [ ] BIPs
- [ ] BuildSource
- [ ] ChangeCoins
- [ ] CoinJoin #1051 
- [ ] ColdWasabi
- [ ] Daemon
- [ ] DeterministicBuild
- [ ] InstallPackage
- [ ] PasswordFinder
- [ ] PayJoin
- [ ] README
- [ ] RPC
- [ ] Receive #1088
- [ ] Send
- [ ] WalletGeneration - @MarnixCroes 
- [ ] WalletRecovery
- [ ] WasabiSetupVM
- [ ] WasabiSetupWhonix

## FAQ
- [ ] Contribution
- [ ] GeneralBitcoinPrivacy
- [ ] Installation
- [x] Introduction #1109 #1112 #1114 #1116 #1120 #1121
- [ ] UseWasabi #1094 #1095 #1096 #1098 #1100 #1101 #1102 #1124 #1125
- [ ] README
  
## GLOSSARY
- [ ] GeneralBitcoin #1122
- [ ] PrivacyWasabi #1119
- [ ] README
  
<br>

- [ ] Update all pictures

<br>

Review and make sure everything works - @Yahiheb

BTCparadigm:
Tasks broken down in to sections. Pick which ever you like and I'll add a user tag behind the task. 
Boxes are checked when that section is updated, MVP-ready and merged.
If some part should be completely removed, open an issue for that discussion.

molnard:
[...]

PR #1091, opened 2022-07-10
[Docs-2.0] Update `Testnet.md` to 2.0

BTCparadigm:
image


MarnixCroes (changes requested):

yahiheb (changes requested):
The pictures should have the default values IMO.
For example `SettingsBitcoin.png` should be replaced where the default dust threshold is shown.
Same for the `SettingsGeneral.PNG` file.

BTCparadigm:
I agree but let's not worry about the settings yet.
https://github.com/zkSNACKs/WasabiDoc/issues/1022#issuecomment-1184515964

yahiheb (approved):
LGTM

PR #1125, opened 2022-08-01
[2.0] Remove faq smaller denomination 0.05

MarnixCroes:


BTCparadigm (commented):
Why delete `What are the fees for the CoinJoin?`?
How about:
```
:::details
### What are the fees for the coinjoin?

Coinjoining coins with a value above 0.01 BTC costs 0.3% as a [coordinator](https://github.com/zkSNACKs/WalletWasabi/tree/master/WalletWasabi.Backend) fee + [mining fees](https://mempool.space/).
Inputs of 0.01 BTC or below don't pay coordinator fees, nor remixes, even after one transaction. 
Thus, a payment made with coinjoined funds allows the sender and the recipient to remix their coins without paying any coordinator fees.
:::

```

MarnixCroes:
> Why delete `What are the fees for the CoinJoin?
 
that was a mistake, fixed.

> How about: ...

done in https://github.com/zkSNACKs/WasabiDoc/pull/1128

yahiheb (approved):

PR #1132, opened 2022-08-03
New Add wallet images 2.0

MarnixCroes:


yahiheb:

MarnixCroes:
added the imgs to the doc

yahiheb (approved):

PR #1133, opened 2022-08-03
Add Coinjoin Strategy img

MarnixCroes:


yahiheb (changes requested):

BTW the icon of this dialog have changed now.

yahiheb (approved):

PR #1134, opened 2022-08-03
add Succes wallet added img

MarnixCroes:


MarnixCroes:
mistake with comment description

PR #1135, opened 2022-08-03
Add AddWalletSuccessfullyAdded img

MarnixCroes:


yahiheb (changes requested):

yahiheb (approved):

PR #1142, opened 2022-08-04
FAQ UseWasabi: update What is dust threshold

MarnixCroes:


docs/FAQ/FAQ-UseWasabi.md
```diff
@@ -1367,8 +1367,8 @@ Dust can mean [a lot of things](https://bitcoin.stackexchange.com/questions/1098
 It can be a non-economical input, that is a UTXO that has less value than the fees it would cost to spend this coin.
 A dust attack is actually about [forced address reuse](https://en.bitcoin.it/Privacy#Forced_address_reuse), the malicious actor sends very small amounts into old addresses and consolidation of these dust UTXOs can link several coins in a wallet cluster.
 
-Specifically in the context of Wasabi, with the dust threshold settings you can limit the value of coins shown in the GUI.
-When you set it to `0.0000 5000 bitcoin`, and when you receive a coin worth `0.0000 4000 bitcoin`, then this coin will not be shown in the coin list.
+Specifically in the context of Wasabi, with the dust threshold settings you can limit the value of coins shown in the GUI as coins *below* the dust threshold are not shown.
```
yahiheb, line 1370:
```suggestion
Specifically in the context of Wasabi, with the dust threshold settings you can limit the value of coins shown in the GUI.
Coins that you receive from other wallets which are *less than or equal to* the dust threshold in value are not shown.
```

docs/FAQ/FAQ-UseWasabi.md
```diff
@@ -1367,8 +1367,8 @@ Dust can mean [a lot of things](https://bitcoin.stackexchange.com/questions/1098
...
+For example: When you set it to `0.0000 5000 BTC`, and when you receive a coin worth `0.0000 4000 BTC`, then this transaction and the coin in the coin list will not be shown.
```
yahiheb, line 1371:
```suggestion
For example: When you set it to `0.0000 5000 BTC`, and when you receive a coin worth `0.0000 4000 BTC`, then this transaction and the coin will not be shown in the coin list.
```

yahiheb (changes requested):

docs/FAQ/FAQ-UseWasabi.md
MarnixCroes: it it got changed for 2.0, as for the 5000 cj outputs to be shown.
so *below* /less than is correct here

yahiheb: If you receive a 5000 sats coin from another wallet it won't be shown.

What has changed/fixed now is that any coin you receive from your own wallet will be shown even if it is 1000 sats, and that is why the 5000 cj outputs are shown.

MarnixCroes: ok I wasn't aware of that thanks

docs/FAQ/FAQ-UseWasabi.md
MarnixCroes: ?

yahiheb (approved):

PR #1148, opened 2022-08-17
updated What is the best time to CoinJoin section

peterchaudhary:
To learn how to make changes, changed CoinJoin to coinjoin.

### DIFF
--- a/docs/FAQ/FAQ-UseWasabi.md
+++ b/docs/FAQ/FAQ-UseWasabi.md
@@ -766,13 +766,13 @@ For example, with a 0.7 BTC input you would expect the following outputs: ~0.1, 
 :::
 
 :::details
-### What is the best time to CoinJoin?
+### When's the best time to coinjoin?
 
-You can CoinJoin whenever you want.
-That said, the ideal way to CoinJoin would be when:
+You can coinjoin whenever you want.
+That said, the ideal way to coinjoin would be when:
 
 1. Mempool is clean.
-CoinJoins are unfairly cheap when the mempool is empty.
+Coinjoins are unfairly cheap when the mempool is empty.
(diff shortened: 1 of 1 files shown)


MarnixCroes (approved):
well done! 
congrats on your first Pull Request

docs/FAQ/FAQ-UseWasabi.md
```diff
@@ -766,13 +766,13 @@ For example, with a 0.7 BTC input you would expect the following outputs: ~0.1,
...
-That said, the ideal way to CoinJoin would be when:
+You can coinjoin whenever you want.
+That said, the ideal way to coinjoin would be when:
 
 1. Mempool is clean.
-CoinJoins are unfairly cheap when the mempool is empty.
+Coinjoins are unfairly cheap when the mempool is empty.
 Remember to take a look at websites like [Johoe's Bitcoin Mempool Statistics](https://jochen-hoenicke.de/queue/#0,24h) or [mempool.space](https://mempool.space/tv) to see mempool statistics in real time.
 
 2. Your inputs are the same or little higher as the minimal denomination for the current round.
```
MaxHillebrand, line 778:
```suggestion
```

yahiheb: I guess this is out of the scope of this PR.

yahiheb (approved):
ACK. Good job @peterchaudhary 

IMO we should remove this whole question as it does not make much sense now given that auto coinjoin is on by default.

peterchaudhary:
@yahiheb we could remove it, but alternatively, since it is an FAQ, we could just change the answer to something along the lines of: "It's best to coinjoin every time you make a transaction. So, Wasabi has made auto-coinjoin the default option...." We can also direct users to how they can opt out of the coinjoin option, and the costs involved (if there are any) when they do coinjoin.

yahiheb:
> we could remove it, but alternatively, since it is an FAQ, we could just change the answer to something along the lines of: "It's best to coinjoin every time you make a transaction. So, Wasabi has made auto-coinjoin the default option...."

ACK

PR #1163, opened 2022-08-31
[2.0] [trivial] FAQUse: update how much anonscore do i need

MarnixCroes:


yahiheb (commented):
What about the answer to this question where anonymity set is mentioned, and the 50 number thrown there?
`It is commonly said that an anonymity set of 50 is sufficient to evade low-level blockchain forensics analysis, but it might not protect you against large adversaries.`

jesseaam:
I am also curious about where this number 50 comes from. 
And do we know if there is a number that will likely protect against large adversaries? Because a question came up the other day on Discord that was getting at this point. 

> Q: What level of anon set would be considered absurd?

Then somebody answered:

>  A:  >250 imo.

<br>

If it's in our documentation, I would like to be able to cite our source - point the user to some additional information that backs up our assertions.

yahiheb:
> I am also curious about where this number 50 comes from.

I don't know.

> > A:  >250 imo.
> 
> 
> If it's in our documentation, I would like to be able to cite our source and point the user to some additional information that backs up our assertions.

No it is not in our documentation AFAIK.

BTCparadigm:
Related to #1218

yahiheb:
@MarnixCroes  Once #1218 is merged, open a new PR to the master branch.

PR #1168, opened 2022-09-01
New step-by-step sending for WW2. Update image.

jesseaam:
I updated the 1st section of the Send page. In particular, I added the new step-by-step for WW2 and included an updated screenshot too.


MarnixCroes (changes requested):
good job
could you put the correct numbers?

docs/using-wasabi/Send.md
```diff
@@ -13,13 +13,15 @@
...
-7. Click `Send Transaction`.
+1. Click the Send tab.
+1. Specify a destination address.
+1. Specify the amount of bitcoin to send to the destination address. Optionally, you can specify the dollar amount to send.
+1. Label the recipient by entering the name of the person or company who you are sending to.
+1. Preview Transaction.<br>
+    a. Verify the amount, recipient, and the address. <br>
+    b. Review the estimated time for confirmation and the transaction fees. Optionally, you can change the transaction fee or confirmation time. <br>
+    c. Review suggestions to improve your privacy by clicking the shield icon in the upper righthand corner. For example, you may want to increase or decrease your payment to avoid change.<br>
+1. Click `Confirm`.
```
yahiheb, line 24: IMO we should add this step:

```suggestion
1. Click `Confirm`.
1. Type in your password, then click `Continue`.
```

yahiheb (changes requested):

yahiheb (changes requested):
IMO the send picture should display the values of the transactions. Discreet mode off.

docs/using-wasabi/Send.md
jesseaam: @yahiheb, are you suggesting that `type in your password` should be a feature added to Wasabi Wallet? Because I haven't seen this prompt in my own experimentation thus far. So adding this to the doc would be premature since the feature has not been added to the wallet yet.

Am I incorrect in my thinking that this feature is not yet in the wallet?

jesseaam:
@MarnixCroes in my latest commit, I switched the format to HTML code so we don't have to worry about the numbering ourselves. Let me know if this is OK, or if it is too inconsistent with legacy format of doc.

@yahiheb, in my next commit I will change the "send picture" to one with discreet mode off so we can see how it looks. I'm also considering adding a collapsible section directly under the "send picture" with pictures of the various steps.

docs/using-wasabi/Send.md
yahiheb: Because you generated a wallet with an empty passphrase/password. Use one with a passphrase.

yahiheb:
> @MarnixCroes in my latest commit, I switched the format to HTML code so we don't have to worry about the numbering ourselves. Let me know if this is OK, or if it is too inconsistent with legacy format of doc.

@jesseaam Yes that is inconsistent with the current format of the doc. I would revert it to how you numbered them the first time.

> could you put the correct numbers?

They will be rendered correctly, so that is fine.

yahiheb (changes requested):
> @MarnixCroes in my latest commit, I switched the format to HTML code so we don't have to worry about the numbering ourselves. Let me know if this is OK, or if it is too inconsistent with legacy format of doc.

@jesseaam Yes that is inconsistent with the current format of the doc. I would revert it to how you numbered them the first time.

> could you put the correct numbers?

@MarnixCroes  They will be rendered correctly, so that is fine.

docs/using-wasabi/Send.md
jesseaam: You're right. I see that now.

yahiheb (approved):
Good job

PR #1194, opened 2022-09-21
npm modifications to get the Docs2.0 to run

jesseaam:
When I cloned the master and tried to deploy it locally, the browser tab was blank when I would visit `http://localhost:8080`.
So then I made a few modifications and it works now. Here is exactly what I did to get it to deploy locally:

```
git clone https://github.com/zkSNACKs/WasabiDoc
cd WasabiDoc
npm install
npm audit fix --force
npm install @vuepress/shared-utils
npm start
```


yahiheb:
@jesseaam Can you test #1195?
I think that PR is better than this one as it updates the dependencies that need to be updated in `package.json` file.

PR #1201, opened 2022-09-26
FAQUse: add privacy bar

MarnixCroes:


docs/FAQ/FAQ-UseWasabi.md
```diff
@@ -198,6 +198,18 @@ It considers the anonymity score weighted amounts instead of just amounts that r
...
+:::details
+### What does the bar with the colored segments mean?
+
+The colored segments make up the `Privacy Bar`.
+The Privacy Bar is supposed to give the user an easy overview of the wallet's coins and it's privacy progress, where each segment represents a wallet's coin.
+The color of each segment indicates if that coin is private (green), semi-private (orange), or non-private (red).
+
+![Wasabi Wallet Privacy Bar](/PrivacyBar.png "Wasabi Wallet Privacy Bar")
+
+So according to the image, this wallet has 46 coins of which 21 are private (green), 11 are semi-private (orange) and 14 are non-private (red).
```
yahiheb, line 210: Maybe this should be removed because the number of those coins is irrelevant because I can't easily count them. WDYT?

yahiheb (changes requested):

docs/FAQ/FAQ-UseWasabi.md
MarnixCroes: correct that it's not easy to count, however I think this surves it's purpose: clarification.
when a user has 10 UTXOs this is doable
maybe in future update the image with lower amount of coins...

yahiheb (approved):

PR #1212, opened 2022-09-28
remove: when best time to coinjoin

MarnixCroes:
quick fix


BTCparadigm (commented):
I'd remove the whole question.

yahiheb (changes requested):
> I'd remove the whole question.

ACK

yahiheb (approved):

PR #1216, opened 2022-09-28
Logo change to light versions

CAnorbo:


MaxHillebrand (changes requested):
the new ones are way smaller, why?

CAnorbo:
> 

Using a 4000x4000 pixel image on a website is not necessary furthermore overkill, so I optimized them for better performance.

yahiheb (approved):

PR #1229, opened 2022-09-29
Send: remove custom change address

MarnixCroes:
remove (for now)
https://github.com/zkSNACKs/WalletWasabi/pull/7108

### DIFF
--- a/docs/using-wasabi/Send.md
+++ b/docs/using-wasabi/Send.md
@@ -124,18 +124,6 @@ If you have a transaction that is high-priority and you really want it to be con
 For a deeper dive into the fee estimation process, [this article](https://bitcointechtalk.com/an-introduction-to-bitcoin-core-fee-estimation-27920880ad0) is worth reading.
 :::
 
-## Custom Change Address
-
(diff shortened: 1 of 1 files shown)


yahiheb (changes requested):
I guess if we remove it then we should also remove https://github.com/zkSNACKs/WasabiDoc/blob/master/docs/FAQ/FAQ-UseWasabi.md#how-can-i-set-a-custom-change-address and `SettingsCustomChange.png`

But we should add these back once it gets implemented.

MarnixCroes:
> I guess if we remove it then we should also remove https://github.com/zkSNACKs/WasabiDoc/blob/master/docs/FAQ/FAQ-UseWasabi.md#how-can-i-set-a-custom-change-address and `SettingsCustomChange.png`
> 
> But we should add these back once it gets implemented.

already a PR, https://github.com/zkSNACKs/WasabiDoc/pull/1230

yahiheb (approved):

yahiheb:
@MarnixCroes When the custom change address is implemented again we should add back this.

PR #1247, opened 2022-10-03
Vuepress 2.x version

CAnorbo:
Updated Vuepress 2.x version.
Fixes #1239  #1238  #1211 #1339 #1342

@yahiheb @MarnixCroes @molnard

PR #1277, opened 2022-10-10
move BuildSource to Advanced Installing

MarnixCroes:
99% of users won't do this, so move it to advanced

### DIFF
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -165,8 +165,7 @@ module.exports = {
           collapsable: false,
           sidebarDepth: 2,
           children: [
-            "/using-wasabi/Download&Install.md",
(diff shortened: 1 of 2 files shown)


MarnixCroes:
changed to draft, wait for after https://github.com/zkSNACKs/WasabiDoc/pull/1247

yahiheb:
NACK, this will break external links.

yahiheb (commented):
NACK, this will break external links.

MarnixCroes:
> NACK, this will break external links.

I think this doesn't break external links?
only the order of which these are displayed in the config and readme

MarnixCroes:
@yahiheb WDYT?

yahiheb (approved):

PR #1284, opened 2022-10-13
first draft of InstallPackage

peterchaudhary:
1. Just saw that the download section of the website was updated to include "/guide" next to "signature". I had not taken this into account, so the format is vastly different and will not be suitable for such an arrangement. I can change the draft to match the requirement if necessary.
2. I need help with the Windows command. I don't have a Windows machine, and it will be great if someone can send me the commands. I'll add them in. They are currently marked with "?"
3. The doc is written for "a mother in her mid 40s or 50s" (as disucssed with Rafe and Max) as the primary audience. I have filled in the gaps of how exactly to verify a signature, and added topics on what GPG and public-key cryptography is.
4. Not sure if and where to add Jesse's videos. I don't have the links, but can surely add them in if required.
5. The DITA framework is loosely followed, where the topics are separated based on their type: concept, reference, and instructional. So, for example, the topic "Install GPG" does not include a brief about what GPG is. That is in another topic "GPG".
6. The main topic of the page is "Install Wasabi," which is why its steps are broken down into prominent sections. Not sure if this format is what you all prefer.
7. Note that I haven't been able to run Vuepress locally, even though Norbo showed me how. So I have used https://dillinger.io/ as my reference. Not sure if the format of the links are correct.
8. Brought in the "Build from source" page, since it is related; the only change made is the heading levels and ordered the headings.


docs/using-wasabi/InstallPackage.md
```diff
@@ -10,147 +10,317 @@
...
-
-	:::tip
-	The output from the verify command may contain `WARNING: This key is not certified with a trusted signature!`.
-	You can ignore this, but if you want to fully verify your download, you need to ask people you trust to confirm that the key fingerprint belongs to zkSNACKs.
+This page provides all the information you will need to install the Wasabi Wallet.
+
+
+### 1. Download installer
+From the [official Wasabi Wallet website](wasabiwallet.io/#download), download the Wasabi installer relevant to your operating system (and processor, in case of MacOS).
+For improved privacy, we recommended that you use the [Tor browser](https://www.torproject.org/), and download the installer from [Wasabi’s official clearnet-website.](http://wasabiukrxmkdgve5kynjztuovbg43uxcbcxn6y2okcrsg7gb6jdmbad.onion/)
```
MarnixCroes, line 18:
```suggestion
For improved privacy, we recommended that you use the [Tor browser](https://www.torproject.org/), and download the installer from [Wasabi’s official clearnet-website.](http://wasabiukrxmkdgve5kynjztuovbg43uxcbcxn6y2okcrsg7gb6jdmbad.onion/)
```

MarnixCroes (commented):
the verification has been clarfied and well explained, I can see that much work has been put into it.
some first comments of mine:

-I would not add `Build from source`, because this is for advanced/developer users, which is a very small amount and causes clutter & confusion for newbs

-personally, I prefer listing the OS and then all the (verification) steps instead of the different OS'es at each step. 
if you think current is better, I'm interested in your argument so please share them

-the images `default-og-image.png` & `download-gpg.png` are not being used?

peterchaudhary:
> -I would not add Build from source, because this is for advanced/developer users, which is a very small amount and causes clutter & confusion for newbs

I think that it reduces clutter in the navigation panel. I'm working with the following outline for `Using Wasabi` section:

- Install Wasabi
- Setup wallet
- Receive bitcoin
- Send bitcoin
- Coinjoin
- Generate wallets
- Connect hardware-wallet
- Connect full-node

So, if we are agreed on the outline, `Build from source` seems to be most suited under "Install Wasabi." 

> -personally, I prefer listing the OS and then all the (verification) steps instead of the different OS'es at each step.

Listing it based on the OS creates a lot of duplicate content, and again, there's a lot of clutter on the page. With tabs, we are going to get a much cleaner output if we arrange content based on the task topic. But even without tabs, it's a lot cleaner and shorter page.

I have updated the page and further trimmed unnecessary content. See http://localhost:3000/en/introduction#install-wasabi-wallet

Plus, I think it helps the user when we arrange topics based on tasks (Install, Verify, Download) rather than on feature or implementation.

> -the images default-og-image.png & download-gpg.png are not being used?
these will be corrected in the next pull request.

I got stuck trying to make a pull request 
image

peterchaudhary:
> > -I would not add Build from source, because this is for advanced/developer users, which is a very small amount and causes clutter & confusion for newbs
> 
> I think that it reduces clutter in the navigation panel. I'm working with the following outline for `Using Wasabi` section:
> 
> * Install Wasabi
> * Setup wallet
> * Receive bitcoin
> * Send bitcoin
> * Coinjoin
> * Generate wallets
> * Connect hardware-wallet
> * Connect full-node
> 
> So, if we are agreed on the outline, `Build from source` seems to be most suited under "Install Wasabi."
> 
> I have also mentioned about it briefly in the overview. We can add more warning tags if requried.
> 
> > -personally, I prefer listing the OS and then all the (verification) steps instead of the different OS'es at each step.
> 
> Listing it based on the OS creates a lot of duplicate content, and again, there's a lot of clutter on the page. With tabs, we are going to get a much cleaner output if we arrange content based on the task topic. But even without tabs, it's a lot cleaner and shorter page. See https://wsb-int-doc.netlify.app/en/introduction/
> 
> I have updated the page and further trimmed unnecessary content.
> 
> Plus, I think it helps the user when we arrange topics based on tasks (Install, Verify, Download) rather than on feature or implementation.
> 
> > -the images default-og-image.png & download-gpg.png are not being used?
> 
> these will be corrected in the next pull request.
> 
> I got stuck trying to make a pull request image

second draft: https://github.com/zkSNACKs/WasabiDoc/commit/117031dcd4f24fb3c34067c3e8754974a4e0e0a0

PR #1303, opened 2022-10-25
change to apt

MaxHillebrand:
`dpkg` is the "wrong" command for installing the package, it does not consider dependency downloads.

`apt` uses `dpkg` but also downloads remote packages.

see https://github.com/zkSNACKs/WalletWasabi/pull/9149 change to apt


MarnixCroes (approved):

BTCparadigm (approved):
2dsag2

yahiheb (approved):

PR #1308, opened 2022-10-25
Install using `apt-get` to allow installing recommended packages (i.e…

kiminuo:
…. `policykit-1`)

Follow-up to https://github.com/zkSNACKs/WalletWasabi/pull/9149#issuecomment-1290225859


Kixunil (approved):

MaxHillebrand:
fwiw I created PR with `apt` here https://github.com/zkSNACKs/WasabiDoc/pull/1303

I only know that `apt-get` is preferred for using with scripts, which we are not doing here.  `apt` is shorter, so I guess that is better.

Kixunil:
Oh, yes `apt` is more user-friendly, it shows progressbar and so on. I don't remember ever coming across a Debian system without it. Thus it makes more sense to prefer it here.

PR #1341, opened 2022-11-09
Fix message when silicon chip uses intel package

turbolay:
Currently, it's wrong.
If I install the Intel package on my m2 it does not crash. It's just slower and uses more ressources because of Rosetta.

Also, can someone tries to run the Wasabi m1 package on an intel mac? Because from what I understand about macOs packages, and according to [this](https://www.maketecheasier.com/run-m1-mac-app-as-intel-app/), it should work as normal.


MarnixCroes (approved):
LGTM

yahiheb (approved):

PR #1351, opened 2022-11-14
how does Wasabi select which coins to send

MarnixCroes:


yahiheb (changes requested):
This is not really correct, it is more nuanced than that.

https://github.com/zkSNACKs/WalletWasabi/blob/master/WalletWasabi.Fluent/ViewModels/Wallets/Send/LabelSelectionViewModel.cs#L63

It is also worth mentioning this:

> Moreover if you have coins that you received in the past from the same recipient (same label as the receiver) Wasabi will use those before even the private coins.

docs/FAQ/FAQ-UseWasabi.md
```diff
@@ -654,6 +654,14 @@ Wasabi will automatically build a transaction with the best combination of the s
 ![Wasabi Wallet Send tab](/Send.png "Wasabi Wallet Send tab")
 :::
 
+:::details
+### How does Wasabi select which coins to send?
+
+Wasabi automatically first selects the private coins.
```
MarnixCroes, line 660:
```suggestion
If the user has a coin with the same label, then Wasabi automatically first selects this coin. 
```

yahiheb (changes requested):

docs/FAQ/FAQ-UseWasabi.md
```diff
@@ -654,6 +654,14 @@ Wasabi will automatically build a transaction with the best combination of the s
 ![Wasabi Wallet Send tab](/Send.png "Wasabi Wallet Send tab")
 :::
 
+:::details
+### How does Wasabi select which coins to send?
+
+Wasabi automatically first selects the private coins. 
+If there are no/not enough private coins for this transaction, then Wasabi will select private coins + non-private coins.
+Which non-private coins will be used can be selected by the user, based on the labelling system.
```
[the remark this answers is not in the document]
MarnixCroes: why? the user selects at "who should know about this transaction" which label to use

docs/FAQ/FAQ-UseWasabi.md
```diff
@@ -654,6 +654,15 @@ Wasabi will automatically build a transaction with the best combination of the s
 ![Wasabi Wallet Send tab](/Send.png "Wasabi Wallet Send tab")
 :::
 
+:::details
+### How does Wasabi select which coins to send?
+
+If there is a coin with the same label (as the recipient label of the outgoing transaction), then Wasabi automatically first selects this coin.
+If this is not the case, or the labelled coin is not enough, then Wasabi will select private coins.
```
yahiheb, line 661:
```suggestion
If the user has coins with the same label (as the recipient label of the outgoing transaction), then Wasabi automatically selects these coins.
If there are no such coins, or they are not enough, then Wasabi will select private and semi private coins if necessary.
```

docs/FAQ/FAQ-UseWasabi.md
```diff
@@ -654,6 +654,15 @@ Wasabi will automatically build a transaction with the best combination of the s
...
+If there are no/not enough private coins for this transaction, then Wasabi will select private coins + non-private coins.
```
yahiheb, line 662:
```suggestion
If there are no/not enough private and semi-private coins for this transaction, then Wasabi will select non-private coins also.
```

yahiheb (changes requested):

yahiheb (changes requested):

docs/FAQ/FAQ-UseWasabi.md
```diff
@@ -654,6 +654,15 @@ Wasabi will automatically build a transaction with the best combination of the s
 ![Wasabi Wallet Send tab](/Send.png "Wasabi Wallet Send tab")
 :::
 
+:::details
+### How does Wasabi select which coins to send?
+
+If the user has coins with the same label as the recipient of the outgoing transaction (coins received in the past from the same recipient), then Wasabi automatically selects these coins.
+If they are not enough, then Wasabi will also select private and semi private coins if necessary.
+If there are not enough private and semi-private coins for this transaction, then Wasabi will select non-private coins also.
+Which non-private coins will be used can be selected by the user, based on the labelling system.
```
yahiheb, line 663:
```suggestion
The user can change which non-private coins will be used, based on the labelling system.
```

yahiheb (changes requested):

PR #1368, opened 2022-11-22
Update max registrable amount

yahiheb:
Merge only if this PR gets merged https://github.com/zkSNACKs/WalletWasabi/pull/9591


MarnixCroes:
> This is not necessary IMO, the docs are general and usually they refer to the latest released version. We are not mentioning what each Wasabi version can and cannot do.

NACK

PR #1422, opened 2023-01-08
Add gap limit tip to resync wallet

yahiheb:


docs/FAQ/FAQ-UseWasabi.md
```diff
@@ -314,6 +314,10 @@ For example if the problem happened 3 days ago then you can go back a week or so
 `new_height = current_height - (7 * 144)`
 :::
 
+:::tip
+If after the synchronization you still think your balance is not correct then increase the [gap limit](/FAQ/FAQ-UseWasabi.html#how-do-i-change-the-gap-limit-of-a-wallet) of the wallet.
```
MarnixCroes, line 318:
```suggestion
If you are doing a re-synchronization, because you expect some missing funds, but after resync you still think your balance is not correct then try to increase the [gap limit](/FAQ/FAQ-UseWasabi.html#how-do-i-change-the-gap-limit-of-a-wallet) of the wallet.
```

MarnixCroes (commented):
the FAQ is more general on how to resync.
not necessarily for missing funds.
I would add that to the tip

cACK

yahiheb:
> the FAQ is more general on how to resync. not necessarily for missing funds. I would add that to the tip

It is a different tip that is why I added it to its own tip.

PR #1423, opened 2023-01-10
Fixes sidebar background on mobile

CAnorbo:


MarnixCroes (approved):

yahiheb (changes requested):
Is this how we want the sidebar to look like?
IMO it is better to have the same background colors if possible.

Capture
Capturess

yahiheb (approved):
Ignore my previous comment it must have been some cache.

tACK


Capture2

PR #1425, opened 2023-01-14
Update ELI5.md

MaxHillebrand:
Makin the ELI5 intro A LOT less complex and with less big words.


yahiheb:
I don't think we need any ELI5 of Wasabi. IMO we should simply change the title of the document to be something like `Wasabi explained simply` or whatever, but we don't need something for real 5 years old.

MarnixCroes:
What's up with this?

PR #1437, opened 2023-01-27
Coins: update manual labeling

MarnixCroes:
quick fix remove WW1 info

the why-wasabi/coins might need a bigger rehash, coinjoin change... etc

### DIFF
--- a/docs/why-wasabi/Coins.md
+++ b/docs/why-wasabi/Coins.md
@@ -32,10 +32,9 @@ This means that coin consolidation can lead to an overall decrease in privacy, e
 
 ### Manual coin labeling and selection
 
-Contrary to many other wallets, Wasabi does not show only the total value of bitcoin in the wallet.
-Rather, in both the `Send` and `CoinJoin` tabs there is a list of all the individual UTXOs.
-Because Wasabi requires users to label every receiving address, the history of each coin is clear for the user at first glance.
-In order to spend a specific coin, it must be manually selected, which prevents the wrong coin from being included in the transaction.
+Using Wasabi it is mandatory to label every receiving and destination address.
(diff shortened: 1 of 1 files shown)


docs/why-wasabi/Coins.md
```diff
@@ -32,10 +32,9 @@ This means that coin consolidation can lead to an overall decrease in privacy, e
 
 ### Manual coin labeling and selection
 
-Contrary to many other wallets, Wasabi does not show only the total value of bitcoin in the wallet.
-Rather, in both the `Send` and `CoinJoin` tabs there is a list of all the individual UTXOs.
-Because Wasabi requires users to label every receiving address, the history of each coin is clear for the user at first glance.
-In order to spend a specific coin, it must be manually selected, which prevents the wrong coin from being included in the transaction.
+Using Wasabi it is mandatory to label every receiving and destination address.
+This helps the user to know where your coins came from so that you can judge whether there are privacy concerns when sending a specific coin to a specific receiver.
```
yahiheb, line 36: I guess since currently the auto coin selection is the default you should be mentioning that the labels help the wallet select the best combination of coins.

```suggestion
This helps the user to know where their coins came from so that they can judge whether there are privacy concerns when sending a specific coin to a specific receiver.
```

yahiheb (changes requested):

docs/why-wasabi/Coins.md
yahiheb: > In Wasabi it is mandatory to give every receive address a label of those who know that this address is yours, so that later you know what to do with it, and it also helps the auto coin selection algorithm when sending.

Something like the above mentioned in https://docs.wasabiwallet.io/using-wasabi/Receive.html#the-importance-of-labeling

yahiheb (approved):

Issue #1452, opened 2023-02-24
Update Restoring Wasabi Wallet in Electrum


yahiheb:
The `Extended Account Private Key zprv` was removed in https://github.com/zkSNACKs/WalletWasabi/pull/9742, because according to Lucas:
> The problem with zprv is that it is for m/84'/xxxxx + segwit only and now we have taproot too.

An example of a user trying to recover with Electrum, but he couldn't because `Extended Account Private Key zprv` is missing from the WW v2.0.2.1 and docs are not up to date. https://github.com/zkSNACKs/WalletWasabi/discussions/10183

We should update the docs accordingly.

Also Lucas mentioned that `there is a tool on electrum that can discover your keys. Honestly I don't remember much but that's would be the way. `
Is this worth investigating how this tool works and mention it in the docs?

MarnixCroes:
> Also Lucas mentioned that there is a tool on electrum that can discover your keys. Honestly I don't remember much but that's would be the way. 
Is this worth investigating how this tool works and mention it in the docs?

there is such a thing, but we should not mention it.

PR #1479, opened 2023-04-04
update RestoreElectrum

MarnixCroes:
closes https://github.com/zkSNACKs/WasabiDoc/issues/1452

-remove recovery using master private key
-remove recovery by manually creating wallet file

not possible anymore


docs/using-wasabi/RestoreElectrum.md
```diff
@@ -31,60 +31,15 @@ An other wallet that does support Taproot should be used for recovering Taproot
 
 3. Choose `Standard wallet`.
 
-4. Choose `I already have a seed` or `Use a master key`.
+4. Choose `I already have a seed`.
 
-5. Type your seed or paste your master private key.
+5. Type in your seed.
```
yahiheb, line 36:
```suggestion
5. Type in your seed (recovery words).
```

yahiheb (changes requested):
IMO we should keep this (at least for a while) for WW1 users and for WW2 version prior to v2.0.2.1.

MarnixCroes:
I think we shouldn't.
We have other places where we decided to have the doc only match the latest release version, like with max coinjoin amount (and not: Wasabi version X do this, Wasabi version X cannot do this etc.).
We should keep it that way.

> Because it is useful for users using any version prior to v2.0.2.

we could have small note and just link to the archive if really needed

yahiheb (approved):

PR #1482, opened 2023-04-06
What do i need to recover wallet + create table

MarnixCroes:
this question still comes up all the time.
we have similar FAQ listed like _Can I recover my wallet without the password?_
but a general one is missing

I think it would be good to have this with a clear table (for real newbs)
image

### DIFF
--- a/docs/FAQ/FAQ-UseWasabi.md
+++ b/docs/FAQ/FAQ-UseWasabi.md
@@ -1072,6 +1072,20 @@ The wallet file backup is sensitive, in terms of privacy, but not critical in te
 So it is good advice to encrypt this wallet file.
 :::
 
+### What do I need to recover my wallet?
+
+To recovery your wallet you need your Password + Recovery words, or Password + Wallet File, as shown in the table below.
+
+| | Recovery Words | Wallet File | Password
+:------------ | :-------------:| :-------------:| :-------------:
+Recovery Words |  |  :x: | :heavy_check_mark:
+Wallet File | :x: |   | :heavy_check_mark:
+Password | :heavy_check_mark: |  :heavy_check_mark: | 
+
+:::tip
+If there was an emtpy password (no password) entered at the wallet creation, then the Recovery Words or Wallet File alone are enough to recover the wallet.
+:::
+
 ### Can I recover my wallet without the password?
 
 No.
--- a/docs/FAQ/FAQ-UseWasabi.md
+++ b/docs/FAQ/FAQ-UseWasabi.md
@@ -1074,7 +1074,7 @@ So it is good advice to encrypt this wallet file.
 
 ### What do I need to recover my wallet?
 
-To recovery your wallet you need your Password + Recovery words, or Password + Wallet File, as shown in the table below.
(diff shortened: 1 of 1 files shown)


docs/FAQ/FAQ-UseWasabi.md
```diff
@@ -1072,11 +1072,25 @@ The wallet file backup is sensitive, in terms of privacy, but not critical in te
 So it is good advice to encrypt this wallet file.
 :::
 
+### What do I need to recover my wallet?
+
+To recovery your wallet you need your Password + Recovery words, or Password + Wallet File, as shown in the table below.
```
yahiheb, line 1077:
```suggestion
To recover your wallet you need either your Recovery Words + Password, or the Wallet File + Password, as shown in the table below.
```

docs/FAQ/FAQ-UseWasabi.md
```diff
@@ -1072,11 +1072,25 @@ The wallet file backup is sensitive, in terms of privacy, but not critical in te
...
+
+| | Recovery Words | Wallet File | Password
+:------------ | :-------------:| :-------------:| :-------------:
+Recovery Words |  |  :x: | :heavy_check_mark:
+Wallet File | :x: |   | :heavy_check_mark:
+Password | :heavy_check_mark: |  :heavy_check_mark: | 
+
+:::tip
+If there was an emtpy password (no password) entered at the wallet creation, then the Recovery Words or Wallet File alone are enough to recover the wallet.
```
yahiheb, line 1086:
```suggestion
If no password was entered at the wallet creation, then the Recovery Words or Wallet File alone are enough to recover the wallet.
```

yahiheb (changes requested):

docs/FAQ/FAQ-UseWasabi.md
```diff
@@ -1072,11 +1072,25 @@ The wallet file backup is sensitive, in terms of privacy, but not critical in te
...
 
+### What do I need to recover my wallet?
+
+To recovery your wallet you need your Password + Recovery words, or Password + Wallet File, as shown in the table below.
+
+| | Recovery Words | Wallet File | Password
+:------------ | :-------------:| :-------------:| :-------------:
+Recovery Words |  |  :x: | :heavy_check_mark:
+Wallet File | :x: |   | :heavy_check_mark:
+Password | :heavy_check_mark: |  :heavy_check_mark: |
```
yahiheb, line 1083: I would add an X to indicate that the recovery words or the wallet file alone is not enough (ofc assuming that a password was used which should be considered the default)

```suggestion
Recovery Words | :x: | :x: | :heavy_check_mark:
Wallet File | :x: | :x: | :heavy_check_mark:
Password | :heavy_check_mark: | :heavy_check_mark: | :x:
```

yahiheb (changes requested):

docs/FAQ/FAQ-UseWasabi.md
MarnixCroes: does that make sense?

yahiheb: For me it does.
What does an empty spot mean?

PR #1495, opened 2023-04-20
Fix some wrong links

turbolay:
Also, I would argue that we should change every link to be relative instead of absolute so they work in localhost


MarnixCroes (approved):

MarnixCroes:
Thanks!

PR #1498, opened 2023-04-23
Fix youtube thumbnails

### LINKED ISSUE #1497: Replace youtube thumbnails by custom images
Sorry @MaxHillebrand, I think that you are really pretty, but this is not serious:

My proposed solution is to make a really simple design featuring Wasabi and Youtube logo and to replace every 
`Watch the video`
by the custom image.

turbolay:
Fixes #1497 

At least as a temporary fix, I think the doc looks much much better that way.
I checked both in light and dark mode, and both look good IMO.


yahiheb (changes requested):
On dark mode sometimes these are not really visible:

image

yahiheb (changes requested):
https://github.com/turbolay/WasabiDoc/blob/fixYtThumbnails/README.md should be updated to remove `hqdefault.jpg` and `maxresdefault.jpg`

README.md
```diff
@@ -103,6 +103,8 @@ And the index of the video of a playlist (e.g. starts the third video):
 
 If the image doesn't appear, that usually happens with older videos, use hqdefault.jpg instead of maxresdefault.jpg
 
+In case that the tumbnail doesn't fit well with the documentation, a default image can be added as custom thumbnail, located at `/Logo_without_text_with_bg_dark_with_yt.png`. Currently, all thumbnails have been set to this default image as a temporary solution because videos are old, sometimes not under our control, and the thumbnails were irrelevant.
```
yahiheb, line 106:
```suggestion
In case that the thumbnail doesn't fit well with the documentation, a default image can be added as custom thumbnail, located at `/Logo_without_text_with_bg_dark_with_yt.png`.
Currently, all thumbnails have been set to this default image as a temporary solution because videos are old, sometimes not under our control, and the thumbnails were irrelevant.
```

yahiheb (changes requested):

yahiheb (approved):
Good job.

PR #1503, opened 2023-04-26
update what is happening out reg phase

MarnixCroes:


MarnixCroes:
current is wrong, this is better, merging

nitpicks welcome in follow up PR

PR #1504, opened 2023-04-26
Coinjoin: clairfy (unused) bitcoin address

MarnixCroes:
clarify
address reuse may occur (by design)
This is not enforced by coordinator


docs/using-wasabi/CoinJoin.md
```diff
@@ -123,7 +123,7 @@ Next, your Wasabi client generates multiple new Tor identities called **Bob**, w
 Bob sends to the Wasabi coordinator:
 
 * An unblinded credential signed by the coordinator
-* A new unused bitcoin address
+* A new (unused) bitcoin address
```
yahiheb, line 126: Why not simply say:

```suggestion
* A bitcoin address
```

PR #1514, opened 2023-05-02
[release] update verify version, CLI: --version

MarnixCroes:


docs/FAQ/FAQ-Installation.md
```diff
@@ -152,6 +152,8 @@ It's always a good idea to run (one of) the latest Wasabi versions.
 
 In the GUI, go to the SearchBar at the top of the screen and click on `About Wasabi`. 
 Here the current version of your Wasabi is displayed.
+You can also verify the version using the command line by executing `wassabee --version`.
```
yahiheb, line 155: Executing this with the current release (2.0.3) runs Wasabi GUI and doesn't return any version.

MarnixCroes: yes because it's not in that release

yahiheb (approved):
utACK

docs/FAQ/FAQ-Installation.md
yahiheb, line 155: This didn't work on Win10, but `wassabeed --version` worked.

MarnixCroes: idk, it works on Linux

yahiheb, line 155:
```suggestion
You can also verify the version using the command line by executing `wassabee --version` or `wassabeed --version` depending on your OS.
```

yahiheb (changes requested):

docs/FAQ/FAQ-Installation.md
yahiheb: I am testing the right version:

image

lontivero: What¿s the errorlevel?

yahiheb: I didn't run the command line with master, but with the released version from this path `C:\Program Files\WasabiWallet`.

yahiheb: > @yahiheb Could you please confirm?

Btw yes I confirm that.

yahiheb: > We should file an issue for this in the WalletWasabi repo and discuss what to do with that.

Yes I already did it before making that comment: https://github.com/zkSNACKs/WalletWasabi/issues/11688

MarnixCroes: what do you mean _depending on your OS_?
I think we should just put a note that `wassabee --version` doesn't work on win.
until it's fixed

yahiheb: > what do you mean depending on your OS?

I have updated my suggestion to be specific to windows.

MarnixCroes: I see what you mean.
It will currently give the same result, but these are two different things and could in the future give different result.
this makes it look like wassabeed is specfically a windows command.

docs/FAQ/FAQ-Installation.md
```diff
@@ -152,6 +152,10 @@ It's always a good idea to run (one of) the latest Wasabi versions.
 
 In the GUI, go to the SearchBar at the top of the screen and click on `About Wasabi`. 
 Here the current version of your Wasabi is displayed.
+
+You can also verify the version using the command line by executing `wassabee --version`, and `wassabeed --version` for the [Daemon](/using-wasabi/Daemon.md) version.
```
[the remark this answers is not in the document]
yahiheb: ACK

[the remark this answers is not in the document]
MarnixCroes: imho I prefer how it is now

this is the "correct" one, telling people to use daemon version command to check the GUI version, because it doesn't work on one OS feels a bit hacky.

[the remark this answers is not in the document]
MarnixCroes: anyway, if you still think we should only mention wassabeed --version, then let's use that and not spend more time on this.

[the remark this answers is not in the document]
yahiheb: The correct one is to use `wassabee --version` but currently there is an issue with this on Windows, so a note or a hack is needed anyway.
I think we should simply go with Kimi's suggestion and update this once the issue is fixed.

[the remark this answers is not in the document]
MarnixCroes: there is a note 
https://github.com/zkSNACKs/WasabiDoc/pull/1514/files#diff-d2d9f0d8f72baaaae52cec35e4602f430e80d72d4c8929f3f11167ff235d8fddR157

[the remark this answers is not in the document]
yahiheb: We can simply say: "To check the version execute X command" which should work for all OSs and move on.
All this PR is for a small addition on how to check the version.

yahiheb (approved):

Issue #1494, opened 2023-04-19
Update rpc


MaxHillebrand:
When merged https://github.com/zkSNACKs/WalletWasabi/pull/10522

PR #1518, opened 2023-05-04
RPC: specify wallet in request path

MarnixCroes:
closes https://github.com/zkSNACKs/WasabiDoc/issues/1494


yahiheb (changes requested):
This tip is valid for all methods that can be used with a wallet so it should be added separately, either at the top or bottom.

Also add an example how to do this, like in https://github.com/zkSNACKs/WalletWasabi/pull/10522#issue-1673583899

MaxHillebrand (commented):
I guess this should be updated consistantly for other calls too

MarnixCroes:
> it should be added separately

pushed

> I guess this should be updated consistantly for other calls too

current is ok? Or do you mean to update all RPC request examples by including the specified wallet?

MaxHillebrand:
> Or do you mean to update all RPC request examples by including the specified wallet?

Yes, can be a follow up pr

yahiheb (approved):
LGTM

Issue #1404, opened 2022-12-16
link not visible in yellow box (cautious warning) on dark mode


MarnixCroes:
![image](https://user-images.githubusercontent.com/93143998/208097725-e4835ba5-ff75-4079-b81f-88a0669dff0a.png)

light mode 
![image](https://user-images.githubusercontent.com/93143998/208097797-f8f1a185-9cb9-46c3-81ce-4e329a210390.png)

yahiheb:
How about simply using `tip` instead of `warning` in such cases.

MarnixCroes:
> How about simply using `tip` instead of `warning` in such cases.

These warnings are useful.
imo we should change the color of the hyperlinks

yahiheb:
> imo we should change the color of the hyperlinks

That might take long time to fix that and I don't see any real difference between `tip` and `warning`.

MarnixCroes:
@CAnorbo 
you might know a quick color fix?

MarnixCroes:
fixed

PR #1525, opened 2023-05-17
Accent color fix

CAnorbo:
Fix for #1404


MarnixCroes (approved):
ACK

yahiheb (dismissed):
This does not fix the issue.

image

yahiheb (changes requested):
@CAnorbo This fixes the issue but first you have to update your branch to master.

yahiheb:
I don't know why but this doesn't seem to fix the issue for me. Maybe because of some cache.

CAnorbo:
> I don't know why but this doesn't seem to fix the issue for me. Maybe because of some cache.

Interesting. If you run **npm start** it should recreate the CSS file automatically, but if I remember well you had the same issue with the previous changes too.

yahiheb:
> > I don't know why but this doesn't seem to fix the issue for me. Maybe because of some cache.
> 
> Interesting. If you run **npm start** it should recreate the CSS file automatically, but if I remember well you had the same issue with the previous changes too.

I don't know, I tried a couple of times and the issue still exists.
@MarnixCroes can you test his PR?

PR #1531, opened 2023-05-31
[release] update CJ output count to 10

MarnixCroes:
https://github.com/zkSNACKs/WalletWasabi/pull/10813

### DIFF
--- a/docs/FAQ/FAQ-UseWasabi.md
+++ b/docs/FAQ/FAQ-UseWasabi.md
@@ -939,7 +939,7 @@ However, as the coordinator cannot enforce this, a custom client could use a dif
 
 ### How many coins/outputs do I get from a coinjoin?
 
-Wasabi Wallet's default client is configured to register 1 up to 8 outputs per coinjoin round.
+Wasabi Wallet's default client is configured to register 1 up to 10 outputs per coinjoin round.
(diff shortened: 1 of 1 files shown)


MarnixCroes:
@MaxHillebrand 
I wasn't aware of that detail, thanks for sharing

yahiheb:
> FYI, in most cases you will register 8 or less, only some edge case whales will register 10.

This is worth mentioning imo.

MaxHillebrand:
> > FYI, in most cases you will register 8 or less, only some edge case whales will register 10.
> 
> This is worth mentioning imo.

Already getting 8 outputs is not likely, getting 9 or 10 has further decreasing likely hood.

I'm not sure if the details of the different internal decomposer have to be explained in this FAQ.

MarnixCroes:
> Already getting 8 outputs has low likely hood, getting 9 or 10 has further decreasing likely hood.
> 
> I'm not sure if the details of the different internal decomposer have to be explained in this FAQ.

Indeed.
This change is not aimed for whales/ has a whale requirement (min x BTC), but it is likely to be used by whales.
i.e. for example in a round there is one whale and multiple plebs, the whale can now further "split" his coins for better privacy/less likely to have a whale output.

yahiheb (approved):

PR #1546, opened 2023-06-21
Add TurboSync entry

turbolay:
I wrote that, I think that the explanation is exhaustive and good, but maybe a bit too long and technical.
Please someone with better english and better skills to write doc (cc @MarnixCroes) it would be great if you can suggest modifications.

### DIFF
--- a/docs/using-wasabi/WalletLoad.md
+++ b/docs/using-wasabi/WalletLoad.md
@@ -59,7 +59,7 @@ In other terms, once an internal key has been used to receive a coin and then th
 Wallets coinjoining a lot will benefit the most from this feature, as the vast majority of their keys will be skipped, reducing the number of blocks to download because of false-positive matches.
 
 Once the wallet is opened, the skipped keys will be tested in the background, in the case that some funds have been received on those keys.
-Users in that case would see their balance update automatically after some time.
+Users in that edge case would see their balance update automatically after some time.
 A log message `Wallet is fully synchronized.` indicates when the verification process finishes.
 
 :::tip You can disable it!
--- a/docs/using-wasabi/WalletLoad.md
+++ b/docs/using-wasabi/WalletLoad.md
@@ -47,7 +47,7 @@ In this step, your Wasabi behaves like any other full node, and cannot be differ
 
 ### TurboSync
 
-Since Wasabi version 2.0.4 a new feature is added called _TurboSync_.
+Since Wasabi version [2.0.4](https://github.com/zkSNACKs/WalletWasabi/releases/tag/v2.0.4)a new feature is added called _TurboSync_.
 _TurboSync_ leads to a significant reduction of wallet loading time, especially for a big wallet.
 
 During a coinjoin, a new address (key) has to be generated for each output.
--- a/docs/using-wasabi/WalletLoad.md
+++ b/docs/using-wasabi/WalletLoad.md
@@ -47,8 +47,8 @@ In this step, your Wasabi behaves like any other full node, and cannot be differ
 
 ### TurboSync
 
-During a Coinjoin, a new address (key) has to be generated for each output.
-As a result, wallets that use the Coinjoin service a lot have a quickly growing set of generated addresses.
+During a coinjoin, a new address (key) has to be generated for each output.
+As a result, wallets that use the coinjoin service a lot have a quickly growing set of generated addresses.
 This is a problem for wallet synchronization: each key has to be tested against each filter, with a small probability of matching as false-positive for every key.
 As a result, addresses with many derived keys will have to download many false positives, making the synchronization more and more time consuming. 
 
--- a/docs/using-wasabi/WalletLoad.md
+++ b/docs/using-wasabi/WalletLoad.md
@@ -50,7 +50,7 @@ In this step, your Wasabi behaves like any other full node, and cannot be differ
 During a coinjoin, a new address (key) has to be generated for each output.
 As a result, wallets that use the coinjoin service a lot have a quickly growing set of generated addresses.
 This is a problem for wallet synchronization: each key has to be tested against each filter, with a small probability of matching as false-positive for every key.
-As a result, addresses with many derived keys will have to download many false positives, making the synchronization more and more time consuming. 
+As a result, wallets with many derived keys will have to download many false-positive blocks, making the synchronization more time consuming.
 
 The TurboSync feature has been built to reduce the hassle caused by synchronization for big wallets.
 It leverages a simple heuristic: Keys used as coinjoin outputs or as change (internal keys) should only be used once.
--- a/docs/using-wasabi/WalletLoad.md
+++ b/docs/using-wasabi/WalletLoad.md
@@ -47,14 +47,22 @@ In this step, your Wasabi behaves like any other full node, and cannot be differ
 
 ### TurboSync
 
-During a Coinjoin, a new address (key) has to be generated for each output. As a result, wallets that use the Coinjoin service a lot have a quickly growing set of generated addresses. This is a problem for wallet synchronization: each key has to be tested against each filter, with a small probability of matching as false-positive for every key. As a result, addresses with many derived keys will have to download many false positives, making the synchronization more and more time consuming. 
+During a Coinjoin, a new address (key) has to be generated for each output.
+As a result, wallets that use the Coinjoin service a lot have a quickly growing set of generated addresses.
+This is a problem for wallet synchronization: each key has to be tested against each filter, with a small probability of matching as false-positive for every key.
+As a result, addresses with many derived keys will have to download many false positives, making the synchronization more and more time consuming. 
 
-The TurboSync feature has been built to reduce the hassle caused by synchronization for big wallets. It leverages a simple heuristic: Keys used as coinjoin outputs or as change (internal keys) should only be used once. In other terms, once an internal key has been used to receive a coin and then this coin was used as input in a new transaction (the coin has been spent), the key should never be used again, and there is no need to test it against higher filters. 
+The TurboSync feature has been built to reduce the hassle caused by synchronization for big wallets.
+It leverages a simple heuristic: Keys used as coinjoin outputs or as change (internal keys) should only be used once.
+In other terms, once an internal key has been used to receive a coin and then this coin was used as input in a new transaction (the coin has been spent), the key should never be used again, and there is no need to test it against higher filters. 
 
 Wallets coinjoining a lot will benefit the most from this feature, as the vast majority of their keys will be skipped, reducing the number of blocks to download because of false-positive matches.
 
-After the wallet will be opened, the skipped keys will then be tested in the background, in the case that some funds have been received on keys skipped. Users in that case would see their balance update automatically after some time. A log message `Wallet is fully synchronized.` indicates when the verification process finishes.
+After the wallet will be opened, the skipped keys will then be tested in the background, in the case that some funds have been received on keys skipped.
+Users in that case would see their balance update automatically after some time.
+A log message `Wallet is fully synchronized.` indicates when the verification process finishes.
 
 :::tip You can disable it!
-TurboSync feature shouldn't cause any problem, but you might want to disable it to debug potential issues with the synchronization of your wallet. In that case, go to your wallet configuration file (search for `Wallet Folder` using Wasabi's searchbar then open the file corresponding to your wallet) and set `UseTurboSync` to `false`
+TurboSync feature shouldn't cause any problem, but you might want to disable it to debug potential issues with the synchronization of your wallet.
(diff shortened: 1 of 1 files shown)


MarnixCroes (commented):
Nice!
Could you make every sentence start on a [new line](https://github.com/zkSNACKs/WasabiDoc/#writing-conventions) plz

docs/using-wasabi/WalletLoad.md
```diff
@@ -44,3 +44,25 @@ When a block filter hits, either a true match or a false positive, then this blo
 If you have [a Bitcoin full node connected](/using-wasabi/BitcoinFullNode.md), then it will fetch the verified block locally.
 If not, then Wasabi will connect to a random Bitcoin P2P node with a new Tor identity, request only this block for download, and then disconnect.
 In this step, your Wasabi behaves like any other full node, and cannot be differentiated.
+
+### TurboSync
+
+During a coinjoin, a new address (key) has to be generated for each output.
+As a result, wallets that use the coinjoin service a lot have a quickly growing set of generated addresses.
+This is a problem for wallet synchronization: each key has to be tested against each filter, with a small probability of matching as false-positive for every key.
+As a result, addresses with many derived keys will have to download many false positives, making the synchronization more and more time consuming.
```
yahiheb, line 53:
```suggestion
As a result, wallets with many derived keys will have to download many false-positive blocks, making the synchronization more time consuming.
```

docs/using-wasabi/WalletLoad.md
```diff
@@ -44,3 +44,25 @@ When a block filter hits, either a true match or a false positive, then this blo
...
+### TurboSync
+
+During a coinjoin, a new address (key) has to be generated for each output.
+As a result, wallets that use the coinjoin service a lot have a quickly growing set of generated addresses.
+This is a problem for wallet synchronization: each key has to be tested against each filter, with a small probability of matching as false-positive for every key.
+As a result, addresses with many derived keys will have to download many false positives, making the synchronization more and more time consuming. 
+
+The TurboSync feature has been built to reduce the hassle caused by synchronization for big wallets.
+It leverages a simple heuristic: Keys used as coinjoin outputs or as change (internal keys) should only be used once.
+In other terms, once an internal key has been used to receive a coin and then this coin was used as input in a new transaction (the coin has been spent), the key should never be used again, and there is no need to test it against higher filters.
```
yahiheb, line 57: I think we can just remove `and then this coin was used as input in a new transaction (the coin has been spent)`.
WDYM by `higher filters`?

```suggestion
In other terms, once an internal key has been used to receive a coin, it should never be used again, and there is no need to test it against higher filters.
```

docs/using-wasabi/WalletLoad.md
```diff
@@ -44,3 +44,25 @@ When a block filter hits, either a true match or a false positive, then this blo
...
+This is a problem for wallet synchronization: each key has to be tested against each filter, with a small probability of matching as false-positive for every key.
+As a result, addresses with many derived keys will have to download many false positives, making the synchronization more and more time consuming. 
+
+The TurboSync feature has been built to reduce the hassle caused by synchronization for big wallets.
+It leverages a simple heuristic: Keys used as coinjoin outputs or as change (internal keys) should only be used once.
+In other terms, once an internal key has been used to receive a coin and then this coin was used as input in a new transaction (the coin has been spent), the key should never be used again, and there is no need to test it against higher filters. 
+
+Wallets coinjoining a lot will benefit the most from this feature, as the vast majority of their keys will be skipped, reducing the number of blocks to download because of false-positive matches.
+
+After the wallet will be opened, the skipped keys will then be tested in the background, in the case that some funds have been received on keys skipped.
```
yahiheb, line 61:
```suggestion
Once the wallet is opened, the skipped keys will be tested in the background, in the case that some funds have been received on those keys.
```

docs/using-wasabi/WalletLoad.md
```diff
@@ -44,3 +44,25 @@ When a block filter hits, either a true match or a false positive, then this blo
...
+Users in that case would see their balance update automatically after some time.
```
yahiheb, line 62: To emphasize that this should be an edge case.

```suggestion
Users in that edge case would see their balance update automatically after some time.
```

docs/using-wasabi/WalletLoad.md
```diff
@@ -44,3 +44,25 @@ When a block filter hits, either a true match or a false positive, then this blo
...
+A log message `Wallet is fully synchronized.` indicates when the verification process finishes.
```
yahiheb, line 63: I think we don't have to mention this.

docs/using-wasabi/WalletLoad.md
```diff
@@ -44,3 +44,25 @@ When a block filter hits, either a true match or a false positive, then this blo
...
+
+:::tip You can disable it!
+TurboSync feature shouldn't cause any problem, but you might want to disable it to debug potential issues with the synchronization of your wallet.
+In that case, go to your wallet configuration file (search for `Wallet Folder` using Wasabi's searchbar then open the file corresponding to your wallet) and set `UseTurboSync` to `false`.
```
yahiheb, line 67:
```suggestion
In that case, go to your wallet file (search for `Wallet Folder` using Wasabi's searchbar then open the file corresponding to your wallet) and set `UseTurboSync` to `false`.
```

yahiheb (changes requested):
This is good.

Should we use the term `address` instead of `key`?

docs/using-wasabi/WalletLoad.md
yahiheb: > This is factually incorect, as the coin will be used again when spending. Do you think we should ignore that?

What if the key was used just to receive, could it be used again? Is there a need to test it against higher filters?

Maybe I am missing something here, but when we talk about reusing an address/key we mean using the address more than once to receive not about spending of course, otherwise there could be no such a thing as address reuse.
IMO yes we can remove it.

> Higher filters I mean higher height filter (so if the address spent at height 1.000, all filters higher than 1000 don't need to be tested against this key). Please make a new suggestion.

I am definitely missing something here :)

docs/using-wasabi/WalletLoad.md
yahiheb: Will this be displayed in the GUI for the user? Plus this is an edge case AFAIK.

yahiheb:
> I introduced the term `key` in the first part of the text. If it was not clear enough for you, it won't be clear enough for users. I will update.

It is clear for me, I am asking if we should make this less technical for readers who might get confused by the term `key`.

docs/using-wasabi/WalletLoad.md
MarnixCroes: We should keep it imo

docs/using-wasabi/WalletLoad.md
```diff
@@ -44,3 +44,25 @@ When a block filter hits, either a true match or a false positive, then this blo
 If you have [a Bitcoin full node connected](/using-wasabi/BitcoinFullNode.md), then it will fetch the verified block locally.
 If not, then Wasabi will connect to a random Bitcoin P2P node with a new Tor identity, request only this block for download, and then disconnect.
 In this step, your Wasabi behaves like any other full node, and cannot be differentiated.
+
+### TurboSync
+
+During a coinjoin, a new address (key) has to be generated for each output.
```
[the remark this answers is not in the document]
MarnixCroes: will the release be 2.0.4 or 2.1?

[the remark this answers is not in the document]
MarnixCroes: I would add this, as it's reach easier if you know what the purpose is of the feature.
And someone who doesn't understand still gets the point.

docs/using-wasabi/WalletLoad.md
MarnixCroes: We should keep it imo

MarnixCroes (commented):
ACK, 
add the correct [release number](https://github.com/zkSNACKs/WasabiDoc/pull/1546/files#diff-7dc9f3c35965e2c1ebe5e083be63f635ff013fdbf4874db2f87360410f6899a5R50) & than merge at release

docs/using-wasabi/WalletLoad.md
yahiheb: Thank you for the explanation, now it is clear.
I would revert the part that says `addresses should be used only twice` to `addresses should be used only once`.
Sorry about the confusion.

docs/using-wasabi/WalletLoad.md
```diff
@@ -44,3 +44,27 @@ When a block filter hits, either a true match or a false positive, then this blo
...
+
+Since Wasabi version 2.x a new feature is added called _TurboSync_.
+_TurboSync_ leads to a significant reduction of wallet loading time, especially for a big wallet.
+
+During a coinjoin, a new address (key) has to be generated for each output.
+As a result, wallets that use the coinjoin service a lot have a quickly growing set of generated addresses.
+This is a problem for wallet synchronization: each address has to be tested against each filter, with a small probability of matching as false-positive for every address.
+As a result, wallets with many derived addresses will have to download many false-positive blocks, making the synchronization more time consuming.
+
+It leverages a simple heuristic: addresses used as coinjoin outputs or as change (internal keys) should only be used twice: one time to receive coins and another time to spend them.
```
yahiheb, line 58:
```suggestion
It leverages a simple heuristic: addresses used as coinjoin outputs or as change (internal keys) should only be used once.
```

docs/using-wasabi/WalletLoad.md
yahiheb: > I am afraid of users stating "Why my balance is incorrect??" because in that case, they should wait until the log message "wallet is fully synchronized"

I do not expect those user to know how TurboSync works and to open the logs file and wait for the message.

I would make it clear that it is a logs message so readers do not expect to see this in the GUI.

```suggestion
Once done the message `Wallet is fully synchronized.` is written to the logs file to indicate that the verification process has finished.
```

yahiheb (changes requested):

docs/using-wasabi/WalletLoad.md
```diff
@@ -44,3 +44,27 @@ When a block filter hits, either a true match or a false positive, then this blo
 If you have [a Bitcoin full node connected](/using-wasabi/BitcoinFullNode.md), then it will fetch the verified block locally.
 If not, then Wasabi will connect to a random Bitcoin P2P node with a new Tor identity, request only this block for download, and then disconnect.
 In this step, your Wasabi behaves like any other full node, and cannot be differentiated.
+
+### TurboSync
+
+Since Wasabi version 2.x a new feature is added called _TurboSync_.
```
MarnixCroes, line 50:
```suggestion
Since Wasabi version 2.0.4 a new feature is added called _TurboSync_.
```

MarnixCroes:
> @yahiheb @MarnixCroes TurboSync has been merged [zkSNACKs/WalletWasabi#10896](https://github.com/zkSNACKs/WalletWasabi/pull/10896) This PR can be merged as well.

nah not yet, as changes get immediately deployed.
Waiting for the release, as trying to keep the docs updated with latest release version

docs/using-wasabi/WalletLoad.md
```diff
@@ -44,3 +44,27 @@ When a block filter hits, either a true match or a false positive, then this blo
 If you have [a Bitcoin full node connected](/using-wasabi/BitcoinFullNode.md), then it will fetch the verified block locally.
 If not, then Wasabi will connect to a random Bitcoin P2P node with a new Tor identity, request only this block for download, and then disconnect.
 In this step, your Wasabi behaves like any other full node, and cannot be differentiated.
+
+### TurboSync
+
+Since Wasabi version 2.0.4 a new feature is added called _TurboSync_.
```
MarnixCroes, line 50:
```suggestion
Since Wasabi version [2.0.4](https://github.com/zkSNACKs/WalletWasabi/releases/tag/v2.0.4)a new feature is added called _TurboSync_.
```

docs/using-wasabi/WalletLoad.md
```diff
@@ -44,3 +44,27 @@ When a block filter hits, either a true match or a false positive, then this blo
 If you have [a Bitcoin full node connected](/using-wasabi/BitcoinFullNode.md), then it will fetch the verified block locally.
 If not, then Wasabi will connect to a random Bitcoin P2P node with a new Tor identity, request only this block for download, and then disconnect.
 In this step, your Wasabi behaves like any other full node, and cannot be differentiated.
+
+### TurboSync
+
+Since Wasabi version [2.0.4](https://github.com/zkSNACKs/WalletWasabi/releases/tag/v2.0.4)a new feature is added called _TurboSync_.
```
MarnixCroes, line 50:
```suggestion
Since Wasabi version [2.0.4](https://github.com/zkSNACKs/WalletWasabi/releases/tag/v2.0.4) a new feature is added called _TurboSync_.
```

MarnixCroes (approved):
Thanks!

PR #1551, opened 2023-06-29
fix external link icon color darkmode

MarnixCroes:
it's not really visible
make it black aswell

image


CAnorbo (changes requested):
It's not working for me, but instead of this line change, in the index.scss file find the html.dark class and add the following to the end:

.external-link-icon {
--external-link-icon-color: none;
}

PR #1572, opened 2023-08-18
 FAQ: coins from Trezor Suite coinjoin not shown 

MarnixCroes:
Until https://github.com/zkSNACKs/WalletWasabi/issues/10802 is resolved, adding this FAQ seems reasonable.

Of course we cannot have explainers for every other wallet (which use different deriv paths etc), but wallets integrating our coinjoins are good to add imo.

### DIFF
--- a/docs/FAQ/FAQ-UseWasabi.md
+++ b/docs/FAQ/FAQ-UseWasabi.md
@@ -1226,7 +1226,8 @@ Thus currently you have to send the bitcoins from your hardware wallet to a `hot
 Read more [here](/using-wasabi/ColdWasabi.md#cold-wasabi-protocol).
 
 :::tip
-Trezor now supports coinjoin with the Trezor Suite, using same rounds as Wasabi Wallet users. Read more [here](https://content.trezor.io/coinjoin)
+Trezor now supports coinjoin with the Trezor Suite, using same rounds as Wasabi Wallet users.
+Read more [here](https://content.trezor.io/coinjoin).
 :::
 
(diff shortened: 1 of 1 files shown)


MaxHillebrand (approved):

MarnixCroes:
I need to add that after changing Taproot path, it needs to rescan

yahiheb (commented):
> To have the coins show up, the `"TaprootAccountKeyPath": "86'/0'/0'"` in the wallet file should be changed to `10025'/0'/0'`.

Have you tested that? I remember doing this in the past and it didn't work.

MarnixCroes:
> > To have the coins show up, the `"TaprootAccountKeyPath": "86'/0'/0'"` in the wallet file should be changed to `10025'/0'/0'`.
> 
> Have you tested that? I remember doing this in the past and it didn't work.

I have not.

MarnixCroes:
Couple users asked this some months ago.
I suggested changing the keypath, iirc it worked.
But if we're not sure, we should not mention it.
So only explain why the coins are not shown, but leave out the suggestion of changing keypath

kravens:
For me the Trezor Suite derivation path is 10025'/0'/0'/1' using SLIP25 and taproot addresses. When I try to edit the wallet.json to this specific path and open the wallet in Wasabi 2.0.4, the wallet never finishes syncing... if I leave out the last /1' it will sync, but not with the same bc1p addresses as in Trezor Suite.

MarnixCroes:
@kravens thank you for ellaborating

yahiheb (approved):

PR #1581, opened 2023-08-23
quick fixes rbf mentions

MarnixCroes:


docs/using-wasabi/ChangeCoins.md
```diff
@@ -52,7 +52,7 @@ This might not be possible in some cases where you have to pay a specific value
 ### Make it difficult to deduce the change
 
 There are [common heuristics identifying change outputs](/why-wasabi/Coins.md#heuristics-identifying-change), try to avoid these with every transaction.
-Don't reuse addresses, don't send precise amounts but randomize them, don't use replace by fee, and try to send to bech32 or bech32m addresses.
+Don't reuse addresses, don't send precise amounts but randomize them, don't use replace by fee (if not necessary), and try to send to bech32 or bech32m addresses.
```
yahiheb, line 55: How about this?

```suggestion
Don't reuse addresses, don't send precise amounts but randomize them, use replace by fee (RBF) only if necessary, and try to send to bech32 or bech32m addresses.
```

yahiheb (changes requested):

PR #1590, opened 2023-09-11
update coin list FAQ

MarnixCroes:
- remove it from `Send` section (it was added there when there was no coin control yet)
- rephrase the question from `how select coins for sending` to `How can I see my wallet's coins`
- kept the image there and mentioning of using it for sending, as that for some cases can still be handy

### DIFF
--- a/docs/FAQ/FAQ-UseWasabi.md
+++ b/docs/FAQ/FAQ-UseWasabi.md
@@ -599,25 +599,6 @@ At the transaction fee slider dialog, click on `Advanced` and manually type the 
 
 ![Wasabi Wallet custom fee rate](/SendCustomFee.png "Wasabi Wallet custom fee rate")
 
-### How do I select coins for spending?
-
-In the normal send workflow, Wasabi automatically selects which coins to send.
-
-To send a specific coin, the user can use the `Wallet Coins` dialog (a.k.a coinlist).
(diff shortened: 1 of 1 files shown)


docs/FAQ/FAQ-UseWasabi.md
```diff
@@ -1429,6 +1410,24 @@ Do NOT use the manual coin control if you do not know what you are doing.
 Misusing it can have seriously bad privacy consequences.
 :::
 
+### How can I see my wallet's coins?
```
yahiheb, line 1413: For seeing coins, the users can use the privacy ring.
The `Wallet Coins` dialog is not for seeing coins, it is for sending specific coins in full.
If we have to rephrase the question, it would be better to change it to:

```suggestion
### How do I select/spend specific coins?
```

yahiheb (changes requested):

docs/FAQ/FAQ-UseWasabi.md
MarnixCroes: privacy ring lacks information, 
like script type, coinjoining or not, confirmation count, unable to see all labels (when many).

yahiheb: Yes it lacks information, but as I mentioned the purpose of `Wallet Coins` `dialog` is not to see coins.

yahiheb: @MarnixCroes What's up with this one?

PR #1593, opened 2023-09-11
add couple youtube video's

MarnixCroes:
two great video's 

this one 
image

and this one
image


MarnixCroes:
cc @Kruwed
any more video's we have?
I'm aware of https://yewtu.be/watch?v=UbOAbXjzBJg
image

But I'm not sure where to add it.

And https://yewtu.be/watch?v=fM5pVMj91D0
image

(maybe create a new one of this, as the setup flow changed)

Kruwed:
> cc @Kruwed any more video's we have? I'm aware of https://yewtu.be/watch?v=UbOAbXjzBJg 
> 
> But I'm not sure where to add it.
> 
> And https://yewtu.be/watch?v=fM5pVMj91D0 
> 
> (maybe create a new one of this, as the setup flow changed)

Those are the most current ones, and yes, it would be worth creating new ones especially once the UI refreshment is done.  Additionally, users' initial deposits will now always remix at least once even on the savings/speed profiles, so the UX has changed slightly as well.

MarnixCroes:
ok, let's stick with the two video's already in the PR then

MaxHillebrand (approved):

yahiheb (approved):
tACK

yahiheb:
I have noticed that but the video is for both how to speed up or cancel transaction, so it made sense to keep it there under both FAQs.

PR #1607, opened 2023-10-12
Running Daemon does nothing for bandwidth?

nopara73:


nopara73:
I see, it's ok to have it here then.
image

PR #1610, opened 2023-10-16
Improve RPC documentation for Windows

### LINKED ISSUE #1608: Create RPC Documentation for Windows
It doesn't work. A person with Windows must go through it and document the necessary steps for Windows.

Related:
- curl setup
- jq setup
- escaping json characters https://github.com/zkSNACKs/WalletWasabi/issues/11678

yahiheb:
Closes #1608

### DIFF
--- a/docs/using-wasabi/RPC.md
+++ b/docs/using-wasabi/RPC.md
@@ -46,6 +46,12 @@ It is recommended to install the `jq` [command line json processor](https://sted
 Then start Wasabi Wallet and load the wallet you want to use.
 Open a new terminal and use the following RPC commands to interact with your wallet, instead of the GUI.
 
+:::tip For Windows Users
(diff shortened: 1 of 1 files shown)


MaxHillebrand (commented):
Maybe add an example?
I don't understand exactly what this means

nopara73 (changes requested):
Bare minimal. I like your thinking.

- [ ] Have you try your suggestions?
- [ ] `curl` is also not default on Windows I think
- [ ] example (what Max said)
- [ ] I haven't tried git bash, have you? Does it really come with `curl` and `jq`? Just confirming.

yahiheb:
> Have you try your suggestions?

Yes ofc.

> `curl` is also not default on Windows I think

Windows 10, version 1803 or later.

> example (what Max said)

Done.

> I haven't tried git bash, have you? Does it really come with `curl` and `jq`? Just confirming.

Yes I always use Bash, it is good.
No it doesn't come with those but if you use Bash you don't need to use double quotation marks or to escape them, you just use the methods as they are now in the docs. IMO this is the best option.

nopara73 (approved):
LGTM utACK

PR #1634, opened 2024-01-09
clarify min input amount

MarnixCroes:
@CAnorbo is working on fixing this https://github.com/zkSNACKs/WasabiStats/issues/20, than I'll add a link where users can see what the actual current amount is.


MarnixCroes:
can add the clarification now and add the link later, or wait for the link

yahiheb (approved):

Issue #1511, opened 2023-04-28
add `Starting Wasabi` (with parameters)


MarnixCroes:
https://github.com/zkSNACKs/WalletWasabi/pull/10293

MarnixCroes:
at Advanced Using Wasabi

PR #1637, opened 2024-01-10
create StartupParameters pillar

MarnixCroes:
as the Start up parameters are not Daemon specific, remove them from the Daemon into a new pillar at Advanced Using Wasabi

closes https://github.com/zkSNACKs/WasabiDoc/issues/1511

### DIFF
--- a/docs/.vuepress/config.ts
+++ b/docs/.vuepress/config.ts
@@ -184,12 +184,12 @@ export default {
                 collapsable: false,
                 sidebarDepth: 2,
                 children: [
+                    "/using-wasabi/StartupParameters.md",
                     "/using-wasabi/Daemon.md",
                     "/using-wasabi/RPC.md",
                     "/using-wasabi/PayJoin.md",
                     "/using-wasabi/Testnet.md",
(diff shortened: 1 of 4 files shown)


MaxHillebrand (dismissed):

docs/.vuepress/config.ts
```diff
@@ -188,7 +188,8 @@ export default {
                     "/using-wasabi/RPC.md",
                     "/using-wasabi/PayJoin.md",
                     "/using-wasabi/Testnet.md",
-                    "/using-wasabi/BIPs.md"
+                    "/using-wasabi/BIPs.md",
+                    "/using-wasabi/StartupParameters.md"
```
yahiheb, line 192: I think it is better to put this as the first sub page before the Daemon.

yahiheb (changes requested):

docs/.vuepress/config.ts
MarnixCroes: done

yahiheb (approved):

PR #1638, opened 2024-01-10
add PlebStopThreshold to glossary

MarnixCroes:


MaxHillebrand (dismissed):

docs/glossary/Glossary-PrivacyWasabi.md
```diff
@@ -152,6 +152,11 @@ The Tor onion service, IP address or domain of the receiver is included in a BIP
 Peers in our documentation refers mainly to Bitcoin and Wasabi Wallet users, but it also means people.
 They are literally peers in the network, or in the CoinJoin.
 
+### Pleb Stop Threshold
+
+The Pleb Stop Threshold is the amount at which the wallet will not automatically start coinjoin, or it will stop coinjoin when the wallet balance is below this amount.
```
yahiheb, line 157: The auto coinjoin won't start if the total wallet balance is less than the plebstop.

```suggestion
The Pleb Stop Threshold is the minimum wallet balance at which coinjoin will automatically start.
```

yahiheb (changes requested):

docs/glossary/Glossary-PrivacyWasabi.md
Kruwed: This is called "Auto-start coinjoin threshold" in the wallet UI, shouldn't the docs match for search purposes even if the internal name is "Pleb Stop"?

MarnixCroes: could. 

As the _auto-start coinjoin threshold_ is in the UI with a tooltip explainer
image
I thought explaining plebstop makes more sense, as that verb gets used sometimes (in marketing?) but it's not explained anywhere

yahiheb: `Auto-start Coinjoin (Pleb Stop) Threshold` is good enough.

yahiheb (approved):

PR #1640, opened 2024-01-11
RPC: excludefromcoinjoin

MarnixCroes:
https://github.com/zkSNACKs/WalletWasabi/pull/11540

third parameter _true_ is not needed, however I think this is the easiest way to write the example in combination with setting it to false.

any objections @turbolay ?


yahiheb (approved):
tACK

Issue #1630, opened 2024-01-05
Changing background color while browsing


molnard:
The background color changes while you are clicking on different topics. There is a vertical gradient from grey to black. At the bottom of the topics, it is completely black. It is not good to read. At least me and chatgpt think like that, so we are two: 

> Reading text on a monitor with a solid black background can be challenging for several reasons:
> 
> Contrast and Glare: High contrast between a solid black background and light-colored text (like white) can cause glare, leading to eye strain. This is because the bright text can scatter more light, creating a halo effect around the words that makes it harder to focus.
> 
> Text Bleeding: On some screens, especially those with lower resolution, white text on a black background can cause a bleeding effect, where the white bleeds into the black, making the text appear less sharp.
> 
> Eye Fatigue: For many people, reading light text on a dark background for prolonged periods can lead to quicker eye fatigue compared to dark text on a light background. This can be due to the way our eyes adjust to the light levels and the effort required to focus on the brighter text.
> 
> Reflections: On glossy screens, a black background can lead to more visible reflections, which can be distracting and make reading difficult.
> 
> Personal Preference and Habituation: Some people may find it harder to read on a black background simply because they are more accustomed to dark text on a light background. This preference can vary widely among individuals.
> 
> It's important to note that these issues can vary depending on the individual's eyesight, the quality and settings of their monitor, and the ambient lighting conditions. Some people may actually find a dark mode (light text on a dark background) more comfortable, especially in low-light environments or if they are sensitive to bright light.

MarnixCroes:
this is the design
What is here to be done, remove the gradient color?

friendly ping @CAnorbo 
what was the motivation to have this?

CAnorbo:
#1646

PR #1646, opened 2024-01-16
Remove gradient background

CAnorbo:
Fixes https://github.com/zkSNACKs/WasabiDoc/issues/1630


MaxHillebrand (approved):

CAnorbo:
> I have tested this and the gradient background is still there. Idk if it is because of some cache.

It's probably cache. @MarnixCroes Do you have it so?

yahiheb (approved):
Dismiss my previous comment, it works.

tACK

PR #1648, opened 2024-01-16
glossary: coinjoin time preference

MarnixCroes:
it is nowhere explained in the docs

the GUI contains a tooltip with this, however some people might not be using the GUI
So good to have it mentioned/explained somewhere

### DIFF
--- a/docs/glossary/Glossary-PrivacyWasabi.md
+++ b/docs/glossary/Glossary-PrivacyWasabi.md
@@ -94,6 +94,12 @@ Read more: [What is a CoinJoin?](/FAQ/FAQ-Introduction.md#what-is-a-coinjoin)
 A _Coinjoin Strategy_ contains instructions for the automatic "coinjoin robot" about configurations like when and how much to coinjoin.
 Read more: [Coinjoin Strategy](/Using-Wasabi/CoinJoin.md#coinjoin-strategy)
 
+### Coinjoin Time Preference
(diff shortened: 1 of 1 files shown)


docs/glossary/Glossary-PrivacyWasabi.md
```diff
@@ -94,6 +94,12 @@ Read more: [What is a CoinJoin?](/FAQ/FAQ-Introduction.md#what-is-a-coinjoin)
 A _Coinjoin Strategy_ contains instructions for the automatic "coinjoin robot" about configurations like when and how much to coinjoin.
 Read more: [Coinjoin Strategy](/Using-Wasabi/CoinJoin.md#coinjoin-strategy)
 
+### Coinjoin Time Preference
+
+The wallet will only participate in coinjoin when the fee rate is below the median of this time frame.
```
yahiheb, line 99: How about this?

```suggestion
The wallet will only participate in coinjoin when the fee rate is below the median of the chosen time frame (Hours, Days, Weeks, or Months).
```

yahiheb (changes requested):

docs/glossary/Glossary-PrivacyWasabi.md
MarnixCroes: done

yahiheb (approved):

PR #1656, opened 2024-01-23
daemon: Remove params from daemon executable

MarnixCroes:
some people think that these params are mandatory to run the daemon.
so remove them, to prevent confusion. better to have only the minimal executable there.

having some params at the examples is good enough.

### DIFF
--- a/docs/using-wasabi/Daemon.md
+++ b/docs/using-wasabi/Daemon.md
@@ -51,7 +51,7 @@ wassabeed
 
 ### If building from source code
 
-Open the terminal, navigate to the WalletWasabi.Daemon directory (inside the cloned repository) and execute the desired commands.
+Open the terminal, navigate to the _WalletWasabi.Daemon_ directory (inside the cloned repository) and execute the desired commands.
 
 ```bash
 $ dotnet run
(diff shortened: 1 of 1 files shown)


docs/using-wasabi/Daemon.md
```diff
@@ -32,29 +32,30 @@ Depending on your operating system, open the command line and execute:
...
+wassabeed
 ```
 
 ### If building from source code
 
 Open the terminal, navigate to the WalletWasabi.Daemon directory (inside the cloned repository) and execute the desired commands.
 
 ```bash
-$ dotnet run --wallet=WalletName --jsonrpcserverenabled=true
+$ cd WalletWasabi/WalletWasabi.Daemon
```
yahiheb, line 57: The text above mentions that already.

```suggestion
```

yahiheb (changes requested):

docs/using-wasabi/Daemon.md
MarnixCroes: correct
I think this doesn't hurt?

yahiheb: I don't know it assumes that you start your terminal where the closed repository is, I would rather not change that. Take a look at the discussions here: https://github.com/zkSNACKs/WasabiDoc/pull/874

yahiheb (approved):

PR #1665, opened 2024-01-30
update taproot faq

MarnixCroes:
At the time, I'm quite sure that someone said that change is always to a taproot address, but it's not.

was probably a misunderstanding or miscommunication.

anyway, fix it here

### DIFF
--- a/docs/FAQ/FAQ-UseWasabi.md
+++ b/docs/FAQ/FAQ-UseWasabi.md
@@ -235,12 +235,10 @@ It also shows some information about the current coinjoin round.
 
 ### Does Wasabi support Taproot?
 
-Partially, Wasabi supports:
-- sending to Taproot (bech32m) address format
-- receiving Taproot outputs from a coinjoin
-- change from a normal transaction is always received to a Taproot address
+Wasabi supports Taproot.
+However, it is currently not possible to generate a Taproot addresses in the GUI.
 
-However, generating a Taproot receive address is not yet supported.
+> While the user cannot generate a Taproot address in the GUI, his wallet might have Taproot coins (from coinjoin, or change output from a single user transaction).
 
 ## Synchronization
 
--- a/docs/FAQ/FAQ-UseWasabi.md
+++ b/docs/FAQ/FAQ-UseWasabi.md
@@ -235,8 +235,7 @@ It also shows some information about the current coinjoin round.
 
 ### Does Wasabi support Taproot?
 
-Wasabi supports Taproot.
-However, it is currently not possible to generate a Taproot address in the GUI.
+Wasabi supports Taproot, except that the user cannot generate Taproot addresses.
 
 > While the user cannot generate a Taproot address in the GUI, his wallet might have Taproot coins (from coinjoin, or change output from a single user transaction).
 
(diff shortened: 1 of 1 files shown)


yahiheb (commented):
I am not sure either, so I let @lontivero decide on this one.

docs/FAQ/FAQ-UseWasabi.md
```diff
@@ -238,7 +238,7 @@ It also shows some information about the current coinjoin round.
 Partially, Wasabi supports:
 - sending to Taproot (bech32m) address format
 - receiving Taproot outputs from a coinjoin
-- change from a normal transaction is always received to a Taproot address
+- receiving Taproot change from a normal transaction
```
lontivero, line 241: I've read this 10 times and have no idea what it means.

MarnixCroes: changed it

docs/FAQ/FAQ-UseWasabi.md
```diff
@@ -238,7 +238,7 @@ It also shows some information about the current coinjoin round.
 Partially, Wasabi supports:
 - sending to Taproot (bech32m) address format
 - receiving Taproot outputs from a coinjoin
```
yahiheb, line 240: As Lucas said I guess this one also should be changed.

MarnixCroes: I understood it as that (only) line 241 didn't make sense

lontivero:
I think we should define what we mean buy support taproot. In the UI there is nothing yes. Internally Wasabi uses taproot for some outputs in the coinjoins (this means that the amount decomposer takes the script types into account) and for change outputs in normal transactions almost randomly too.
In the coin selector IDK but in the RPC we can see the taproot scripts/addresses, we can generate new taproot addresses (so, we can receive money from others to those taprrot addresses) , etc. We can pay in coinjoin to a taproot script. We generate compact filters with taproot scripts in order to recognize those blocks containing transactions from/to our wallet, etc,etc

yahiheb (commented):
I would simplify this by saying:

Wasabi supports taproot.
- Sending to taproot addresses is supported.
- Generating taproot addresses is supported using the RPC but not in the UI.

A note can be added explaining more how this works for coinoins.

MarnixCroes:
simplified it by saying Wasabi supports Taproot, except for address generation in GUI.

and put a small note to mention possibility of having taproot coins.

MarnixCroes:
> Generating taproot addresses is supported using the RPC but not in the UI.

the `getnewaddress` RPC doesn't return taproot, so this can be confusing

lontivero:
Sorry but I think it is still confusing. I would say that Wasabi supports taproot except for that fact that it doesn't allow you to create taproot addresses. I you don't agree then merge it as it is.

MarnixCroes:
> Sorry but I think it is still confusing. I would say that Wasabi supports taproot except for that fact that it doesn't allow you to create taproot addresses. I you don't agree then merge it as it is.

ok done: _Wasabi supports Taproot, except that the user cannot generate Taproot addresses._

yahiheb (approved):

Issue #1612, opened 2023-11-07
Search in content


turbolay:
Currently, the search bar only searches in the titles, it would be much better to search in the content as well.
I talked to @CAnorbo about it a few months ago and if I remember correctly there should be an option just to change this behavior.

CAnorbo:
For this, there is a 3rd party search engine called Algolia that we could use with VuePress.
First, we have to send our website for indexing and they'll give us an API key.

Read more here: https://v2.vuepress.vuejs.org/reference/plugin/docsearch.html

turbolay:
Can we proceed with this? I think it would be a really nice improvement. I've seen it's free and it seems easy to implement

MaxHillebrand:
We're already using algolia, afaik it searches content too.

turbolay:
> We're already using algolia, afaik it searches content too.

We are only searching in the titles right now. 
In our doc, ctrl c + d a line from the doc and you won't have result
In this doc https://v2.vuepress.vuejs.org/reference/plugin/docsearch.html#indexname, ctrl c + d a line will find the relevant article. It's much better. I'm having a really hard time navigating the doc because of this.

yahiheb:
Fixed by #1613.

yahiheb:
Reopening this since #1613 was reverted temporarily by #1617

yahiheb:
Again reopening this since #1622 was reverted by #1625

molnard:
👁️

MarnixCroes:
cACK

@CAnorbo any plan for this? -> steps what to do next?

CAnorbo:
We have two options, the old search or the new. Choose one and go with it.
It's a really basic framework, if we'd like to keep it, there are no other options.

MarnixCroes:
ok. 
the new search would be great, if it works correctly.
It was removed because some basic searches were missing/not working.

CAnorbo:
It's a scheduled crawler., that runs I guess daily.

molnard:
@yahiheb What is the status of this?

yahiheb:
There is nothing new about it, @CAnorbo tried to fix this using Algolia but couldn't get it to work properly so we reverted the changes.

molnard:
Maybe there was a misunderstanding. What is the next step @yahiheb ? wontfix?

yahiheb:
[...]

PR #1668, opened 2024-02-15
Update Vite version to latest

CAnorbo:
Solves server vulnerability issues.


yahiheb:
The number and the order of results has changed (maybe not a big deal) when searching for something after this PR:

Before:

SR0

After:

SR1

Should we do something about it, or we just wait until a proper fix for #1612 is made?

PR #1672, opened 2024-02-16
Change search engine

CAnorbo:
- removed old search engine
- added search pro

### DIFF
--- a/docs/.vuepress/config.ts
+++ b/docs/.vuepress/config.ts
@@ -3,7 +3,7 @@ import { viteBundler } from '@vuepress/bundler-vite'
 import { defineUserConfig } from 'vuepress'
 import defaultTheme from '@vuepress/theme-default'
 import slugify from '@vuepress/shared-utils'
-import searchPlugin from '@vuepress/plugin-search'
+import { searchProPlugin } from "vuepress-plugin-search-pro";
 import palettePlugin from '@vuepress/plugin-palette'
 
 
@@ -267,15 +267,11 @@ export default defineUserConfig({
                 }
             }
         }],
-        searchPlugin({
-            // getExtraFields: (page) => page.frontmatter.tags,
-            maxSuggestions: 15,
-            hotKeys: ['s', '/'],
-            locales: {
-                '/': {
-                    placeholder: 'Search...',
-                }
-            }
+        searchProPlugin({
+          indexContent: true,
+          autoSuggestions: false,
+          searchDelay: 500,
+          sortStrategy: 'total' // 'max' or 'total'
         }),
         palettePlugin({
           preset: 'sass'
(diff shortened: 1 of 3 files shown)


molnard (approved):
LGTM

MarnixCroes (approved):
Let's give it a shot

yahiheb (commented):
I don't think this is currently good enough.

Compare the results when searching for `How do I`:

PR:
N1

Master:
N2

Compare the results when searching for `How do I re`:

PR (same result):
N1

Master:
N3

yahiheb:
Is the following fair enough search for you?

Results for searching the word `coinjoin`:

PR (current master):

C1

Previous master:

C2

Results for searching the word `What is coinjoin`:

PR (current master):

N1

Previous master:

N2

yahiheb:
> You're also not being fair with your searches, people should not use a search engine that way, but instead write a keyword.

That's what technical people do, they know what they are looking for and they know exactly what words to write to get the best results, not the average person though.

Clearly the results above show that this solution as it is not is not close to be good enough, so we don't have to rush to change things and merge this, we already did twice and ended up reverting those changes because we thought they were good enough.
I am not against the solution but I am against rushing when it is not needed and nothing is broken.

The docs has worked as it is for 5 years without a single user complaining about it.
Are we trying to improve this for imaginary users that didn't complain about this or for ourselves and devs?
Technical people can for example simply clone the repository and use `git grep` or other commands to find whatever they need easily.

MarnixCroes:
all links give a 404
see link left corner
image

yahiheb:
Did you guys even click on any results?

This is what I get  https://docs.wasabiwallet.io/undefined.html

CAnorbo:
Something haven't built properly during the deployment. It's working well on local.

molnard:
These are the logs for the deployment. 

https://github.com/zkSNACKs/WasabiDoc/actions/runs/7932241796/job/21658275599

CAnorbo:
Can we check a deployment to the .co without the cache on the Azure Pipeline?
This is the only difference that I can see. I don't have any cache on my computer.

molnard:
I don't know what is .co - can you tell me? 

What is cached? The packages? Maybe this? https://stackoverflow.com/a/74487454

CAnorbo:
> I don't know what is .co - can you tell me? 
> 
> 
> 
> What is cached? The packages? Maybe this? https://stackoverflow.com/a/74487454

.co should be a test version of the docs as we have it on our main website.

On the Azure deploy script I see cache if we can clone that deploy script and test it without cache in a test docs environment it’d be really useful now.

molnard:
> .co should be a test version of the docs as we have it on our main website.

Alright where can I take a look at it?

molnard:
What if we remove the cache part from the script and see what happens?

PR #1674, opened 2024-02-19
add explanation to startcoinjoinsweep

MarnixCroes:
Add some explanation.
&
also put a note that this is **not** a proper coinjoin to other wallet implementation, as some people were already trying this (and assuming it is)

### DIFF
--- a/docs/using-wasabi/RPC.md
+++ b/docs/using-wasabi/RPC.md
@@ -622,6 +622,11 @@ The first parameter is the wallet password, the second parameter is `stopWhenAll
 
 ### startcoinjoinsweep
 
+Sweeps (empties) the wallet by sending the coins in coinjoin to the destination wallet.
+It works the same as normal coinjoin, except that the outputs are sent to (internal addresses of) the destination wallet.
+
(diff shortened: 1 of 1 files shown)


docs/using-wasabi/RPC.md
```diff
@@ -622,6 +622,11 @@ The first parameter is the wallet password, the second parameter is `stopWhenAll
 
 ### startcoinjoinsweep
 
+Sweeps (empties) the wallet by sending the coins in coinjoin to the destination wallet.
+It works the same as normal coinjoin, except that the outputs are send to (internal address(es) of) the destination wallet.
```
yahiheb, line 626: Most likely the coins will go to multiple addresses.

```suggestion
It works the same as normal coinjoin, except that the outputs are sent to (internal addresses of) the destination wallet.
```

yahiheb (changes requested):

docs/using-wasabi/RPC.md
MarnixCroes: done

yahiheb (approved):

PR #1675, opened 2024-02-20
FAQ: add explanation how to exludefromcoinjoin

MarnixCroes:


MarnixCroes:
idk if we should add a FAQ _How can I exclude coins from coinjoin_ (maybe better to wait until this is possible for GUI users)
or rephrasing this FAQ to that

yahiheb (commented):
I think we better add a separate FAQ for this and mention that it is currently possible only using the RPC, so that each FAQ answers a specific question.
Also I don't think we should mention the option to manually add coins to the PrisonedCoins.json file.

PR #1682, opened 2024-03-08
clarify/mention min input count for fresh and blame round

MarnixCroes:


MarnixCroes:
friendly ping @turbolay

MaxHillebrand (dismissed):

MarnixCroes:
> Friendly ACK
> 
> BTW, it would be better to use variables for this intstead of hard coded values, to avoid having to track down everything if we change them

done

kristapsk (approved):

docs/using-wasabi/CoinJoin.md
```diff
@@ -97,7 +97,7 @@ The Wasabi coordinator now verifies that:
 Only when all these checks are passed, does the coordinator allow this input to be registered.
 The coordinator creates and sends a credential back to Alice, which has the same amount as the input minus fees.
 
-The input registration phase ends when either: the number of registered inputs reaches the maximum, or when the time elapsed and the minimum number of inputs is satisfied.
+The input registration phase ends when either: the number of registered inputs reaches the maximum, or when the time elapsed and the minimum number of inputs (${minInputCountFreshRound}) is satisfied.
```
yahiheb, line 100: There is an input registration phase for blame rounds as well.

I would first add this under L78:
`(${minInputCountFreshRound}) for fresh rounds or (${minInputCountBlameRound}) for blame rounds.`

then change this to:

```suggestion
The input registration phase ends when either: the number of registered inputs reaches the maximum, or when the time elapsed and the minimum number of inputs is satisfied, ${minInputCountFreshRound} or ${minInputCountBlameRound}.
```

yahiheb (changes requested):
Min input is mentioned here also, update it:

https://docs.wasabiwallet.io/FAQ/FAQ-UseWasabi.html#what-is-happening-in-the-input-registration-phase

Should we also add a variable for the max input count?

docs/using-wasabi/CoinJoin.md
yahiheb: Is there anything wrong in the above suggestion?
It just describes accurately what happens/could happen in the input registration phase.

yahiheb: A blame round is not a phase by itself required for a successful coinjoin, it is just a coinjoin round with all phases (fallback mechanism).

The suggestion doesn't break the flow and there is nothing saying that this step-by-step is specific for fresh rounds.
I would rather be precise and accurate when describing these phases specially with blame rounds given that many fresh rounds fail.

yahiheb (approved):

PR #1685, opened 2024-03-10
Update CoinJoin.md

zarei673mo:


yahiheb:
The current link is correct, and the proposed change is wrong.

PR #1689, opened 2024-03-18
why wasabi: update change avoidance to privacy suggestions

MarnixCroes:
when change avoidance was [added](https://github.com/zkSNACKs/WasabiDoc/pull/1449), we didn't have the privacy suggestions yet, so update it

### DIFF
--- a/docs/why-wasabi/Coins.md
+++ b/docs/why-wasabi/Coins.md
@@ -38,12 +38,14 @@ The default _Send_ workflow uses the auto coin selection algorithm, which is bas
 
 Read more [here](/using-wasabi/Receive.md#the-importance-of-labeling).
 
-### Change avoidance suggestion
+### Privacy suggestions
 
-When sending bitcoin, the transaction might contain change (back to the sender).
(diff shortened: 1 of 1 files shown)


yahiheb (commented):
Why removing the `Change avoidance suggestion` part?
I think we could keep it and add the `Privacy suggestions` separately.

MarnixCroes:
> Why removing the Change avoidance suggestion part?

because the change avoidance is a privacy suggestion

yahiheb:
So because of that it should be removed as a solution to avoid change?

Does `Privacy suggestions` fit in this document about coins?

MarnixCroes:
privacy suggestions are about coins:
don't select non-private coins, only select semi-private coins, private coins, don't have change coins, select coins of same label, selecting more than 10 coins, using coinjoining coins etc

yahiheb:
But out of all those cases change avoidance stands out, and thus [my suggestion](https://github.com/zkSNACKs/WasabiDoc/pull/1689#pullrequestreview-1944238714) to keep it and add the `Privacy suggestions` separately.
One other reason to keep it is to not break any external links unless really necessary.

PR #1691, opened 2024-03-21
Switch Electrum to Sparrow to include Taproot (P2TR) 

Gustavojfe:
We need to present a method to users on how to recover their wallet with another wallet.

So far, we had an article on our docs page for Electrum, but Electrum doesn't support P2TR addresses.

After taking a look at WalletsRecovery.org, it's clear that Sparrow is the only wallet that can recover P2TR + passphrase.

So I updated this Docs page to switch Electrum to Sparrow.

If a user follows this guide, he'll be able to recover for both P2WPKH and P2TR.


MarnixCroes (commented):
thanks for the suggestion

It is correct that recovering in Electrum will not show taproot funds, however we discussed this some time ago an it was decided to keep it like this.
iirc it was because some would argue we shouldn't spend time on, and have a document about other wallets at all, but we decided to keep Electrum as it doesn't hurt.

tbh I'm not sure if we should change it, it will break links (and maybe Electrum will support taproot soon?)

anyway, these would need to be updated too

- https://github.com/zkSNACKs/WasabiDoc/blob/2fe7207df1632cc3dbd048f1e51366bdb3653633/docs/.vuepress/config.ts#L176
- https://github.com/zkSNACKs/WasabiDoc/blob/2fe7207df1632cc3dbd048f1e51366bdb3653633/README.md?plain=1#L131

yahiheb:
As Marnix mentioned above this was discussed before and the decision was not to add any guides on how to recover funds in any other wallet.

The derivation paths that Wasabi uses are already mentioned and that should help users who are looking to recover their funds elsewhere:

https://docs.wasabiwallet.io/FAQ/FAQ-UseWasabi.html#what-derivation-paths-does-wasabi-use
https://docs.wasabiwallet.io/FAQ/FAQ-UseWasabi.html#does-wasabi-support-taproot

MaxHillebrand:
> decision was not to add any guides on how to recover funds in any other wallet.

Are you sure? Why?

yahiheb:
Yes I am.
Because if something is broken with Wasabi we should work on fixing it not on suggesting our users to move their bitcoin to other wallets, specially competitor ones like Sparrow, and specially that most/all other wallets by default do not have the privacy guarantees that Wasabi has.
Also we shouldn't take the responsibility to tell the users which wallets to use, we just need to provide enough information about the standards Wasabi uses to allow them to recover elsewhere whenever they please.

Gustavojfe:
I had a call with Rafe and Max around 10 days ago and we discussed this and I was told to do this.

It's unfortunate that it's a competitor, but Sparrow is the only wallet that allows a user to recover Wasabi P2TR addresses with a passphrase.

I strongly doubt this will make us lose users, and I don't see it as a recommendation, but as a necessity to users that are not technically savvy to recover from derivation paths. 

Wasabi is not a common wallet, it requires a connection to the backend. Some users have difficulty connecting to the backend, and it doesn't just block them from coinjoins, but from accessing their funds.

lontivero:
When we released the very first version of wasabi we wanted to let the users know that their bitcoin were not in an incompatible wallet, that they could recover their wallets in any other wallet if some catastrophic event would happen to wasabi.

That was to help new users to come and trust on a very new, and at the time futuristic, bitcoin wallet.

Nothing catastrophic had happened and now we know that our response time under any kind of emergency is pretty good.

We didn't write a how-to recover in Electrum because we like Electrum as we shouldn't write a how-to move to any other wallet regardless of how awesome/disastrous those other wallets are.

Other wallets can write their own compatibility check lists and how-tos. I suggest to do that in Wasabi too. I suggest to remove the how-to recover in Electrum completely and add a compatibility list and maybe a section about how to recover with Wasabi seed phrases created with other wallets.

Perpetuate something that could have made some little sense in the past is not okay, instead, imagine that how-to never existed and someone come with a "what if we write a how to recover wasabi wallets in superwow wallet!?"

Gustavojfe:
that makes total sense to me, if all agree I can take care of that

yahiheb:
> add a compatibility list and maybe a section about how to recover with Wasabi seed phrases created with other wallets.

ACK

Gustavojfe:
OK I'll close this and open a new PR

PR #1741, opened 2024-06-12
Update more UI images

yahiheb:


MarnixCroes (approved):
cack
is it on purpose to also increase all the image sizes 2 a 3 times?

yahiheb:
> is it on purpose to also increase all the image sizes 2 a 3 times?

No, that's the size I get them when I take the screenshots.
In fact I don't like that some of them are really big, we can work on improving all of them later on I guess.

PR #1752, opened 2024-06-27
FAQ contribution: update how to donate

MarnixCroes:


turbolay (approved):
We will update this again when all channels will be created

PR #1760, opened 2024-07-03
add latest coordinator configs to glossary

MarnixCroes:
closes https://github.com/WalletWasabi/WasabiDoc/issues/1755 https://github.com/WalletWasabi/WasabiDoc/issues/1738 https://github.com/WalletWasabi/WasabiDoc/issues/1736

For GUI users the tooltip is self explanatory.
Adding it here to the glossary for non GUI users and a good way to have it at least mentioned somewhere in the docs, including the default values.


docs/glossary/Glossary-PrivacyWasabi.md
```diff
@@ -14,6 +14,11 @@ In the case of Wasabi documentation, it usually indicates the arrival of a new f
 
 E.g. "Lightning Network is coming to Wasabi in #twoweeks"
 
+### Absolute min input count
```
yahiheb, line 17: How about changing this to `Min input count` and put it with the other coordinator configs?

docs/glossary/Glossary-PrivacyWasabi.md
```diff
@@ -147,6 +152,16 @@ KYC (Know Your Customer) is the process of a business being forced to identify a
 The term is also used to refer to the bank regulation which governs these activities.
 Read more: [AML/KYC Information](/why-wasabi/TransactionSurveillanceCompanies.md#aml-kyc-information)
 
+### Max Coordination Fee Rate
+
+Maximum coordination fee rate the client is willing to accept to participate into a round.
+The default value is 0.0.
```
yahiheb, line 158:
```suggestion
The maximum coordination fee rate the client is willing to accept to participate into a round.
The default value is 0%.
```

docs/glossary/Glossary-PrivacyWasabi.md
```diff
@@ -147,6 +152,16 @@ KYC (Know Your Customer) is the process of a business being forced to identify a
...
+
+### Max Coinjoin Mining Fee Rate
+
+Maximum mining fee rate in sat/vByte the client is willing to pay to participate into a round.
+The default value is 150.
```
yahiheb, line 163:
```suggestion
The maximum mining fee rate in sat/vByte the client is willing to pay to participate into a round.
The default value is 150 s/vb.
```

yahiheb (changes requested):

docs/glossary/Glossary-PrivacyWasabi.md
MarnixCroes: adding them here is aimed for daemon users. 
`AbsoluteMinInputCount` is the actual name of the property, so I prefer this.
for GUI users with the tooltip it should be self-explanatory

yahiheb (approved):

lontivero:
@MarnixCroes this documentation contains a common problem that confused rates with percentages. Yes, rates can be expressed as percentages when the relation is between two magnitudes of the same kind but it would be preferable to use rates everywhere and avoid percentages.

MarnixCroes:
> @MarnixCroes this documentation contains a common problem that confused rates with percentages. Yes, rates can be expressed as percentages when the relation is between two magnitudes of the same kind but it would be preferable to use rates everywhere and avoid percentages.

fixed in https://github.com/WalletWasabi/WasabiDoc/pull/1775

PR #1764, opened 2024-07-04
Warn to use only well reviewed password recovery tools

yahiheb:
I made this PR mainly to test #1763 to fix the docs deployment problem.


lontivero:
Imo all that section makes no sense. We should says that password is required and if we want to recommend something we should recommend a password manager or something like that. Cracking a decent password requires special hardware and it is in practice almost unfeasible

PR #1789, opened 2024-07-17
Update Sybil attack sentence

yahiheb:
Remove mention of zkSNACKs coordinator.


MarnixCroes:
The FAQ needs to be rewritten in full https://github.com/WalletWasabi/WasabiDoc/issues/1779 that's why I didn't remove it previously

PR #1813, opened 2024-09-07
Update WalletGeneration.md

watou:
I noticed the ASCII flowchart and I thought I'd take a crack at converting it to mermaid.

### DIFF
--- a/docs/using-wasabi/WalletGeneration.md
+++ b/docs/using-wasabi/WalletGeneration.md
@@ -80,46 +80,23 @@ This means that the passphrase is your last line of defense against anyone who t
 
 Wasabi uses [BIP 38: Password-Protected Private Key](https://github.com/bitcoin/bips/blob/master/bip-0038.mediawiki)
 
-```
-                      +--------------+
-                      | Entropy      |
-                      +--------------+
-  +------------+             |
-  | Word list  +------------>+
-  +------------+             |
-                      +------v-------+
-                      | Mnemonics    |
-                      +--------------+
-                             |
-                   +--------->
-                   |         |
-                   |  +------v-------+
-                   |  | Seed         |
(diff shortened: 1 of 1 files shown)


MarnixCroes (commented):
hi,

Concept ACK, your current proposal doesn't render the chart though

### master
image

### pr
Screenshot from 2024-09-07 18-59-33

MarnixCroes:
it seems there is a vuepress pluging that would make it work
That plugin would need to be added to the config, I'm not sure if it's worth is tbh.

How about the adding mermaid chart as an image?
Screenshot from 2024-09-07 19-32-37

watou:
I didn't mean to give unwanted work to anyone!  I'm just fond of viewing mermaid charts over ASCII versions -- less cognitive friction IMHO.

I won't take offence if you don't find it worthwhile.  All the best!

MarnixCroes:
@watou 
no worries!
Your proposal to use a mermaid chart is an improvement.
If you could add it as an image then I'm happy to add it this change.

MarnixCroes:
the image should be added to /docs/vuepress/public.

the mentioning of `this is needed for bip38` is removed now.
I think it is still good to mention it, maybe add it like this:
image


```
pass[Passphrase] ----> jctB 
pass[Passphrase] -- This is needed to use BIP38 -->  jctC
```
The arrows consolidating in the small dots seems to be unnecessary.
Maybe remove the dots and just link them to the boxes? WDYT?

thanks

watou:
Hopefully last commits are good!

yahiheb (changes requested):
Why is the `Passphrase` at the top? I would rather see it aligned with the `Mnemonics`.

I am not sure `This step is needed to use bip38` is at the right place.

PR #1815, opened 2024-09-11
add back Wasabist to public announcement websites

MarnixCroes:
It was previously removed as they were recommending a specific coordinator (their own).
They don't do that anymore.

So they can be added back


MarnixCroes:
@turbolay gave ACK in DM, merging

PR #1822, opened 2024-09-16
update default max amount

MarnixCroes:
https://github.com/WalletWasabi/WalletWasabi/pull/13403/


turbolay:
but we will re revert it

PR #1836, opened 2024-09-26
Auto-update version number

turbolay:
This PR adds an action that is triggered from the main repository. The action takes as parameter a version number, then goes to config.ts and automatically replaces the version number.

Because of branch protections it cannot commit directly to master so instead it commits on a new branch, creates a PR, merges it with admin privileges.

For some reason, it is hard for an action to trigger another action, even with a merge like this. So the deploy action is then triggered through the api after the merge.

This PR works on my fork, maybe it will have some kind of permission issues here. It only uses the auto-generated access token from github

### DIFF
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -4,6 +4,8 @@ on:
   push:
     branches:
       - master
+  repository_dispatch:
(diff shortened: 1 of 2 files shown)


yahiheb (approved):
cACK

PR #1841, opened 2024-10-10
coinjoin: add leftover going to coordinator

MarnixCroes:
Mention/explain leftovers going to the coordinator.
It is currently nowhere mentioned in the docs.

image


MarnixCroes:
I'm not sure if we should mention that it happens "rarely"/ is usually 0.
as it depends on curcumstances

turbolay:
Please write 10.000 and not 5.000, reason being that the coordinator can set this limit but the client will only accept 10.000.

I would rewrite something like that:
In rare cases, up to 10.000 sats per coinjoin can go to the coordinator. This is because getting these small amounts back would harm privacy and end up being more expansive than just forfeiting it.

MarnixCroes:
@turbolay applied your suggestion and rewrote your suggested text a bit.
please check

PR #1846, opened 2024-10-20
SEND: update fee estimation

MarnixCroes:


MarnixCroes:
@lontivero @turbolay please correct me if I'm wrong, I did not check the code.

PR #1854, opened 2024-11-09
remove offline Wasabist

MarnixCroes:
They have been offline for several weeks, seems like it's not a temporary issue

happy to add them back if they come online again


MarnixCroes:
@turbolay and I agreed last week to have them listed a little longer. waiting for if they come back online, but they didn't
so, merging now

PR #1858, opened 2024-11-19
[Trivial] Use secure link (https)

yahiheb:


MarnixCroes:
FTR
https wasn't working for me, http was
they're just shitty server where it's running on, sometimes it works sometimes not

PR #1859, opened 2024-11-21
update Can the coordinator attack me FAQ

MarnixCroes:
closes https://github.com/WalletWasabi/WasabiDoc/issues/1779

### DIFF
--- a/docs/FAQ/FAQ-Introduction.md
+++ b/docs/FAQ/FAQ-Introduction.md
@@ -101,19 +101,41 @@ Also, remember to follow our [blog](https://blog.wasabiwallet.io) to get the lat
 
 ### Can the coordinator attack me?
 
-The developers have gone to great lengths to ensure that the coordinator cannot steal funds nor link inputs to outputs. 
-The nature of Wasabi is that you should not need to trust the developers or the Wasabi coordinating server, as you can verify that the code does not leak information to anyone.
-Due to the nature of coinjoin transactions users don't need to trust other users or the coordinator against theft, leaving denial of service and attacks on privacy as the main concerns
+Wasabi is designed with a zero trust policy.
+This also applies to the coordinator: it doesn't need to be trusted as it cannot steal funds or link inputs to outputs.
+All computation, including decomposition, is done on the client side.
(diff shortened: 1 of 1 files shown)


MarnixCroes:
@lontivero @turbolay do you mind reviewing, and is there anything I'm forgetting?

Kruwed:
Should it be mentioned that disconnecting during input registration leaks metadata to the coordinator if multiple inputs disappear at the same time? Users should avoid pausing their coinjoin near the end of the input registration timer to avoid this sort of timing analysis.

MarnixCroes:
> Should it be mentioned that disconnecting during input registration leaks metadata to the coordinator if multiple inputs disappear at the same time? Users should avoid pausing their coinjoin near the end of the input registration timer to avoid this sort of timing analysis.

hmm, it's a good thing to mention somewhere in the docs, but maybe not in this FAQ.

In this case the de-registration is done because of an action from the client, it's not an attack that a coordinator can "always" do and Wasabi has a defense built in for.
So keeping this FAQ narrow to possible attacks from the coordinator would be my preference.

or maybe adding it as a small side note in this FAQ would be good, but I wouldn't list it as the four other points.

MarnixCroes:
> or maybe adding it as a small side note in this FAQ would be good, but I wouldn't list it as the four other points.

done 505937bc319c0aece057ba34a715776ac4f5894d

see images in PR description how it looks now

turbolay:
Can you check this and take some idea and leave out what you don't like then please merge without waiting for approval https://github.com/WalletWasabi/WasabiDoc/commit/7caf281f6ec73fa9606347a29810d1af07d010ba

MarnixCroes:
> Can you check this and take some idea and leave out what you don't like then please merge without waiting for approval [7caf281](https://github.com/WalletWasabi/WasabiDoc/commit/7caf281f6ec73fa9606347a29810d1af07d010ba)

Thank you for the great improvement.
I've applied it and made some minor tweaks (start sentence on new line, wording tweaks etc)

MarnixCroes:
Good to go!

turbolay:
Auto deploy doesn't work again??? grrrrr

PR #1882, opened 2025-01-20
SEND: update change avoidance percentage

MarnixCroes:
https://github.com/WalletWasabi/WalletWasabi/pull/13666

### DIFF
--- a/docs/using-wasabi/Send.md
+++ b/docs/using-wasabi/Send.md
@@ -142,7 +142,7 @@ There are clickable suggestions to:
 - Not use non-private coins (only use private and semi-private)
 - Avoid change
 
-The suggestion to only use semi-private or private coins is only shown if the amount difference is less than or equal to 25%.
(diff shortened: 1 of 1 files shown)


docs/using-wasabi/Send.md
```diff
@@ -142,7 +142,7 @@ There are clickable suggestions to:
 - Not use non-private coins (only use private and semi-private)
 - Avoid change
 
-The suggestion to only use semi-private or private coins is only shown if the amount difference is less than or equal to 25%.
+The suggestion to only use semi-private or private coins is only shown if the amount difference is less than or equal to 15%.
```
yahiheb, line 145: We can add change avoidance here since we don't mention this information elsewhere which we probably should.

```suggestion
The suggestions to only use private or semi-private coins are shown only if the amount decreases at most by 15%.
The suggestion to avoid change is shown only if the amount increases/decreases at most by 15%.
```

yahiheb (changes requested):

yahiheb (approved):

PR #1949, opened 2025-05-06
update Full node wrt RPC

MarnixCroes:


MarnixCroes:
merging is it's much better than current one.
improvements can be done in follow ups

PR #1956, opened 2025-05-27
Searchable SLIP39 support

yahiheb:
To make SLIP39 reachable from the search bar.

Another option is create a new document for supported SLIPs but it's not worth it.


yahiheb:
Currently the SLIP39 support subsection in the `BIPs.md` document is not searchable because you can only have one main title/section per document.

So it still makes sense to fix that.

PR #1969, opened 2025-07-23
startup params: move one item down to exactly match config file

MarnixCroes:


MarnixCroes:
to match the order of the items

PR #1971, opened 2025-07-24
update wasabi's solution: block filters over tor

MarnixCroes:


MarnixCroes:
clear improvement over current, which is wrong.
merging

PR #1977, opened 2025-08-21
Update website download images

MarnixCroes:


MarnixCroes:
ran locally: the measurements of the new images are ok

PR #1990, opened 2025-09-24
Fix grammar in Bitcoin Privacy FAQ

Kruwed:
Small fixes. Also removes the run-on sentence `Only in a perfectly enlightened discrimination-free world where no one has undue authority over anyone else could we retain our dignity and make our lawful transactions freely without self-censorship if we don't have privacy.`


MarnixCroes (approved):
ACK, thanks

PR #2009, opened 2025-10-31
update FAQ: Outgoing transactions assume full RBF

Kruwed:
Transactions no longer signal the BIP125 flag. I don't know why the final line in the doc is edited.


MarnixCroes (approved):
good catch.
Should we explicitly mention that transactions don't signal RBF?
It's nowhere mentioned in the docs.

Kruwed:
>Should we explicitly mention that transactions don't signal RBF? It's nowhere mentioned in the docs.

I don't think so, the story behind the BIP125 flag is a 10 year long battle.

PR #2051, opened 2026-03-10
add signet support

### LINKED ISSUE #2039: signet support
https://github.com/WalletWasabi/WalletWasabi/pull/14183

MarnixCroes:
closes https://github.com/WalletWasabi/wasabidoc/issues/2039


MarnixCroes:
> There is the dedicated `Testnet` https://docs.wasabiwallet.io/using-wasabi/Testnet.html, chapter. Maybe later this can be expanded into a more general test network document. Adding a dedicated `signet` chapter is overkill as it's basically the same as testnet, but for RegTest we could maybe add one.

adding it to testnet chapter in follow up

PR #2064, opened 2026-04-02
Scheme scripting first entry

### LINKED ISSUE #2030: scheme scripting language
https://github.com/WalletWasabi/WalletWasabi/pull/14042

MarnixCroes:
closes https://github.com/WalletWasabi/WasabiDoc/issues/2030

image

### DIFF
--- a/docs/using-wasabi/SchemeScripting.md
+++ b/docs/using-wasabi/SchemeScripting.md
@@ -9,6 +9,12 @@
 
 [[toc]]
 
+:::danger
+This is an experimental feature.
+Only execute commands you understand.
+Commands can drain your funds and/or compromise your privacy.
+:::
+
 ## What it is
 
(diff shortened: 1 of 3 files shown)


docs/using-wasabi/SchemeScripting.md
```diff
@@ -0,0 +1,50 @@
...
+---
+
+# Scheme Scripting language
+
+[[toc]]
+
+:::danger
+This is an experimental feature.
+Only execute commands you understand.
+Commands can drain your funds and/or compromise your privacy.
```
lontivero: I think this is not true, or at least not yet, because the scripts don't have access to keys, network nor file system.

I would say that it is not tested enough and the possibilities are not completely understood. So, we don't know what can be done and what can not.

MarnixCroes: removed line 15, kept the others.
Especially `Only execute commands you understand.` covers everything.

PR #2073, opened 2026-05-18
Fix build

lontivero:
The site was not working because there were deep dependency incompatibilities caused by the stupid dependantbot which suggests updating individual dependencies while ignoring that there exist inter-dependencies.

Here an example, imagine there are only two packages (package A and package B) as follow:

* Package A v1.0.0
* Package B v1.0.1 (depends on A v1.0.0)

dependantbot detects that package A v1.0.0 has a vulnerability which was fixed in v1.0.17 and creates a PR to upgrade package A so, the resulting packages will be:

* Package A v1.0.17
* Package B v1.0.1 (which still depends on A v1.0.0 booom!!!)

@MarnixCroes when merging one of those PRs, it is important to check the build passed. In case it fails we need to revert it. Here the problems is that this was merged in March but there were other PR changing dependencies since then; none of them worked.

### DIFF
--- a/docs/.vuepress/config.ts
+++ b/docs/.vuepress/config.ts
@@ -1,10 +1,10 @@
 import resolve from 'path'
 import { viteBundler } from '@vuepress/bundler-vite'
 import { defineUserConfig } from 'vuepress'
-import defaultTheme from '@vuepress/theme-default'
+import { defaultTheme } from '@vuepress/theme-default'
 import slugify from '@vuepress/shared-utils'
-import searchPlugin from '@vuepress/plugin-search'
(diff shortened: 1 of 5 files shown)


MarnixCroes:
> @MarnixCroes when merging one of those PRs, it is important to check the build passed. In case it fails we need to revert it. Here the problems is that this was merged in March but there were other PR changing dependencies since then; none of them worked.

I apologize.

I don't pay attention to the build at a PR, because it never works?
Screenshot from 2026-05-18 21-11-04

But I do always run dependabot deps build upgrades locally to see if it still works, and I comment it on those PR's.
I will check tomorrow at which PR it went wrong.

PR #2096, opened 2026-07-23
remove very old note about taproot outputs from coinjoin

MarnixCroes:


MarnixCroes:
added back part of it, makes sense to have as a note.
It is still valid AFAIK, 50% taproot outputs

PR #2099, opened 2026-07-23
Wallet load: update filter download from P2P network

MarnixCroes:


MarnixCroes:
merging as clear improvement.

If there is something, please comment and I'll fix in follow up

PR #2100, opened 2026-07-23
update network level privacy: new P2P sync

MarnixCroes:


MarnixCroes:
merging as clear improvement.

If there is something, please comment and I'll fix in follow up

